PR #5, opened 2025-06-30
Resolution to Issue 3, IPAM and iptable rules' solution

cmyers-mieweb:
**Solution for Issue 3 that does the following.**
1. Copies over a present copy of the nginx port map from the load balancer into the pve1 temp directory
2. Compiles a `pct list` of intern-phxdc-pve1 and intern-phxdc-pve2 and records current hostnames of systems.
3. Checks for containers not present within the pct list and removes iptables rules for hostnames' correlated IP address and ports, freeing them for a new container. 
4. Logs all results to `/var/log/prune_iptables.log`
Test results posted below. 
Screenshot 2025-06-30 143435
image

**Additional Considerations**
* Removing the DHCP lease was also considered, however after inspecting the dnsmasq lease file, it appears that dnsmasq already handles this. I am guessing that the permanent reservation will check whether or not the container is still active and reassign the address with more lease time. Otherwise reserve for when that MAC address re-enters the online state. 
* A cronjob still needs to be configured after this script is reviewed.

### DIFF
--- a/intern-phxdc-pve1/prune_iptables.sh
+++ b/intern-phxdc-pve1/prune_iptables.sh
@@ -1,0 +1,72 @@
+#!/bin/bash
+
+# Script to prune iptables rules for containers that no longer exist
+# It fetches the port_map.json file from a remote host, checks for existing containers, and removes iptables rules for containers that are not present in the current Proxmox setup
+# Location: /usr/local/bin/prune_iptables.sh
+# Author: Carter Myers
+set -euo pipefail
+
+# Remote source of port_map.json
+REMOTE_HOST="intern-nginx"
+REMOTE_FILE="/etc/nginx/port_map.json"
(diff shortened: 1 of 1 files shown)


cmyers-mieweb:
Fixes that have been requested have been implemented in the recent commit. Detailed logging has also been added to illustrate specific steps in the prune process to better show the script's behavior and functions. Verification has also been added to ensure that the port_map.json file is correct and stable to match the exact environment present at the time of script execution. If iptables rules have already been removed, manually, the script will identify this and notify in the logs that it could not locate an iptable rule for the given address tied to the removed container. 

prprune4
prprune3
prprune2
prprune1

maxklema (approved):
Tested the script and it successfully pruned iptable rules for no-longer existing containers. Good to merge PR.

PR #8, opened 2025-07-28
CI/CD GitHub Actions Automation, Helper Scripts, Container Creation Updates, Bug Fixes, Etc.

maxklema:


container creation/get-deployment-details.sh
```diff
@@ -0,0 +1,269 @@
...
+
+if [ "$PROJECT_BRANCH" == "" ]; then
+    PROJECT_BRANCH="main"
+fi
+
+REPOSITORY_BRANCH_EXISTS=$(curl -s -o /dev/null -w "%{http_code}" $PROJECT_REPOSITORY/tree/$PROJECT_BRANCH)
+while [ "$REPOSITORY_BRANCH_EXISTS" != "200" ]; do
+    echo "⚠️ The branch you provided, \"$PROJECT_BRANCH\", does not exist on repository at \"$PROJECT_REPOSITORY\"."
+    read -p "🪾  Enter the project branch to deploy from (leave blank for \"main\") → " PROJECT_BRANCH
+    if [ "PROJECT_BRANCH" == "" ]; then
```
cmyers-mieweb, line 49: This original conditional looks to be making a comparison with a a literal string, perhaps for better accuracy we can check the actual _value_ of the `$PROJECT_BRANCH` var.
```suggestion
   if [ -z "$PROJECT_BRANCH" ]; then
```

container creation/get-deployment-details.sh
```diff
@@ -0,0 +1,269 @@
...
+        fi
+    done
+    UNIQUE_COMPONENTS+=("$COMPONENT")
+}
+
+source /root/bin/deployment-scripts/gatherSetupCommands.sh # Function to gather build, install, and start commands
+
+source /root/bin/deployment-scripts/gatherEnvVars.sh # Gather Environment Variables
+gatherSetupCommands "BUILD" "🏗️  Enter the build command (leave blank if no build command) →  " # Gather Build Command(s)
+gatherSetupCommands "INSTALL" "📦 Enter the install command (e.g., 'npm install') →  " # Gather Install Command(s)echo "$INSTALL_COMMAND"
```
cmyers-mieweb, line 128: Going to guess that the addition of `echo "$INSTALL_COMMAND"` is a redundant comment addition? If it is intentional it can be kept, but marking this for script cleanup.

container maintenance/helper-scripts/verify_container_ownership.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
```
cmyers-mieweb, line 1: I wonder with the addition of the `user` json argument in the container creation process if it will be more efficient to just iterate through `port_map.json` rather than rely on proxmox tags and checking the pct list which also may include containers with multiple tags? May also reduce this down to 1 or 2 jq commands to just grab the container hostname and user field.

container maintenance/update-container.sh
```diff
@@ -0,0 +1,242 @@
...
+
+if [ "$PROJECT_ROOT" == "" ] || [ "$PROJECT_ROOT" == "/" ]; then
+    PROJECT_ROOT="."
+fi
+
+# Install Services ====
+
+echo "🛎️ Installing Services..."
+
+# SERVICE_COMMANDS=$(ssh -o SendEnv="LINUX_DISTRIBUTION SERVICES CUSTOM_SERVICES REQUIRE_SERVICES" \
```
cmyers-mieweb, line 91: does this commented block have an ongoing purpose in the script? Or was it only used during the development cycle of making this?

cmyers-mieweb (changes requested):
Left a couple notes, mostly clarification points or ideas before merge. We can start with these and investigate more further after.

container creation/create-container.sh
```diff
@@ -0,0 +1,199 @@
...
+GREEN='\033[32m'
+RESET='\033[0m'
+
+# Run cleanup commands in case script is interrupted
+
+cleanup()
+{
+
+	echo -e "${BOLD}━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━${RESET}"
+	echo "⚠️  Script was abruptly exited. Running cleanup tasks."
```
runleveldev, line 17: Not all terminals support UTF-8 emojis

container creation/get-deployment-details.sh
maxklema: I agree. I will re-factor these comparisons with empty strings to use the empty flag, -z.

container creation/get-deployment-details.sh
maxklema: Correct. It will be removed.

container maintenance/helper-scripts/verify_container_ownership.sh
maxklema: I like this idea. Prevents having to determine which Hypervisor the container belongs on and using conditional statements. Will refactor.

container maintenance/update-container.sh
maxklema: Yes. This block is uncommented now and fully implemented.

maxklema:
I added a README to each folder and edited the existing README at the root directory. I also used AI to create a mermaid graph of how our cluster works at a high level. I thought it did a good job.

Issue #60, opened 2025-10-14
create-a-container: add a /logout route


runleveldev:
**Issue:** Add `/logout` endpoint to handle user sign-out

**Description:**
There is currently no route to terminate user sessions. Users remain logged in until their session expires manually.

**Recommendation:**
Implement a `/logout` route that:

1. Destroys the active session
2. Redirects the user to the login page

**Example:**

```js
app.get('/logout', (req, res) => {
  req.session.destroy(() => res.redirect('/login'));
});
```

**Benefits:**

* Provides a clear and secure logout mechanism
* Prevents unauthorized session reuse
* Improves user experience and session management compliance

Issue #58, opened 2025-10-14
create-a-container: use authentication checking middleware


runleveldev:
**Issue:** Centralize authentication logic using middleware and add login redirect support

**Description:**
Endpoints currently perform authentication checks manually, e.g.:

```js
if (!req.session.user) {
```

This duplicates logic across routes and makes maintenance error-prone.

**Recommendation:**

* Implement a shared authentication middleware that validates the session and either continues or redirects to the login page.
* Add a `redirect` query parameter to the login page so users are returned to their original destination after successful authentication.

**Benefits:**

* Reduces code duplication and improves consistency
* Simplifies adding new protected routes
* Provides smoother user experience with post-login redirects

Issue #53, opened 2025-10-14
create-a-container: Depend directly on Authentication and do not invoke as cli


runleveldev:
**Issue:** Replace CLI invocation of Node script with direct module import and function call

**Description:**
The current implementation executes a Node.js script as a separate process:

```js
const runner = spawn('node', ['/root/bin/js/runner.js', 'authenticateUser', username, password]);
```

This approach treats a JavaScript file as a command-line tool rather than a reusable module.

**Recommendation:**
Refactor `runner.js` to export its functions and import it directly where needed, for example:

```js
const { authenticateUser } = require('/root/bin/js/runner');
authenticateUser(username, password);
```

**Benefits:**

* Eliminates unnecessary process spawning overhead
* Improves performance and error handling
* Enables direct access to function results and exceptions
* Simplifies testing and maintenance

PR #62, opened 2025-10-14
Reduce external dependencies and improve RESTfulness of create-a-container

### LINKED ISSUE #50: create-a-container: use extensionless urls
**Issue:** Use semantic, extensionless routes instead of file-based ones

**Description:**
Route handlers like https://github.com/mieweb/opensource-server/blob/412d65d428d4079a956342ee80bee64f71f3c272/create-a-container/server.js#L56 should be updated to use clean, semantic paths.

**Example:**
Replace:

```js
app.get('/form.html', (req, res) => {
```

With:

```js
app.get('/containers/new', (req, res) => {
```

**Rationale:**

* Removes implementation details from URLs
* Improves security by obscuring file structure
* Simplifies future refactoring and routing changes

runleveldev:
Closes #50, #53, #58, #60

As is, these changes are hard to test due to the dependencies on the create-a-container server environment, but I did my best to replicate it at https://create-a-container-dev.opensource.mieweb.org/ for demonstration. The UI is fully functional but I did not replicate all scripts necessary to actually create a container. Functionality should be similar enough to avoid breaking via the UI however.

### DIFF
--- a/create-a-container/views/form.html
+++ b/create-a-container/views/form.html
@@ -64,7 +64,7 @@
         const formData = new FormData(this);
         const data = Object.fromEntries(formData.entries());
 
-            try {
+        try {
             const res = await fetch('/containers', {
                 method: 'POST',
                 headers: { 'Content-Type': 'application/json' },


cmyers-mieweb (approved):
Looks good, going to perform test against create-a-container prod to deploy and test as a prelude to the new opensource node.

PR #72, opened 2025-10-14
add missing files and updated README

runleveldev:
I want to throw out this mass of BASH and replace it with an actual system using pure NodeJS handled by create-a-container. All "frontends" (ssh, GH Actions, etc.) should just call back to the /containers/new API. We especially need to eliminate the mass of nested-ssh calls as root to other servers. We then need to audit authorized keys because I'm afraid there's been a lot of over-provisioning for permissions. In particular, all calls to `pct` and family should be replaced by calls to the Proxmox API and use the user's credentials instead of root's.

This PR does document what currently exists, and also re-organizes everything to match what's currently in-use on the cluster. The README will also include instructions on how to deploy changes to these files while we implement the replacements.


cmyers-mieweb (approved):
Looks Good, we will set this as  a new baseline

Issue #75, opened 2025-10-20
CSS Issues present in the create-a-container request account page


cmyers-mieweb:
Some buttons such as the "Request Account" and fields such as "email" and "date" do not have any CSS elements applied to them.

PR #76, opened 2025-10-20
Add styles for sign up button and section

cmyers-mieweb:
Added CSS rules for #signUpBtn and .signup-section to style the sign up button and provide spacing for the sign up section. Includes hover effects and button appearance enhancements. Keeping the back to login button the same to keep it distinct from the actual request. 

Fixes https://github.com/mieweb/opensource-server/issues/75

image

image


runleveldev (approved):
LGTM

Issue #63, opened 2025-10-14
create-container: Update Containers on an interval


runleveldev:
**Issue:** Automate container template updates and pre-apply fungible configuration

**Description:**
Container templates are currently static and require manual updates or reconfiguration. This can lead to outdated configurations and unnecessary manual intervention.

**Recommendation:**

* Implement automatic synchronization of container templates from their upstream sources on a defined interval.
* Pre-apply fungible configuration values (e.g., LDAP settings, default environment variables) that do not depend on runtime-specific details such as hostname or IP address.

**Benefits:**

* Ensures templates remain current with upstream improvements and security patches
* Reduces manual maintenance workload
* Enables faster, more consistent provisioning
* Improves reliability of automated deployments

runleveldev:
See ChatGPT link for the strategy I want to employ which will cover both this issue and #35 
https://chatgpt.com/share/68f95622-d220-8004-9901-d64c5536db80

runleveldev:
This will best be implemented as a 2 part PR. First implement the Hashicorp Packer build process and automation, then once the builds are tested and working in main, update the container creation process to build a new container from that template rather than cloning the given LXC ID.

PR #80, opened 2025-10-24
DRAFT: Add Packer build and upload workflow for Proxmox templates

cmyers-mieweb:
Introduces a draft GitHub Actions workflow to support https://github.com/mieweb/opensource-server/issues/63 to build and upload Debian 12 and Rocky 9 LXC templates for Proxmox. Adds supporting Packer HCL files, Ansible provisioning playbook, Python API utilities for uploading templates, and variable files for template customization.

### DIFF
--- a/.github/workflows/build-templates.yml
+++ b/.github/workflows/build-templates.yml
@@ -1,31 +1,29 @@
 name: Build and Upload Templates
 on:
   push:
-    # Only trigger on pushes that touch files in this packer tree
     paths:
       - 'packer/**'
   schedule:
-    - cron: "0 4 * * *" # Nightly at 4 AM
+    - cron: "0 4 * * *"
   workflow_dispatch:
 
 jobs:
-
   build:
     runs-on: ubuntu-latest
     env:
       PROXMOX_API_URL: ${{ secrets.PROXMOX_API_URL }}
       PROXMOX_TOKEN_ID: ${{ secrets.PROXMOX_TOKEN_ID }}
       PROXMOX_TOKEN_SECRET: ${{ secrets.PROXMOX_TOKEN_SECRET }}
-      # Define the template version ONCE for the whole job
-      TEMPLATE_VERSION: ${{ format('{0:yyyyMMdd}', github.run_started_at) }}
 
     steps:
       - uses: actions/checkout@v4
 
+      - name: Set TEMPLATE_VERSION
+        run: echo "TEMPLATE_VERSION=$(date +%Y%m%d)" >> $GITHUB_ENV
+
       - name: Install dependencies
         run: sudo apt-get update && sudo apt-get install -y packer ansible zstd python3-requests
 
-      # --- Build Debian 12 ---
       - name: Build Debian 12 Template
         run: |
           packer build \
@@ -37,7 +35,6 @@ jobs:
           python3 api/proxmox_upload.py \
             --file /tmp/output/debian12-fungible_${TEMPLATE_VERSION}.tar.xz
 
-      # --- Build Rocky 9 ---
       - name: Build Rocky 9 Template
         run: |
           packer build \


container-creation/intern-phxdc-pve1/packer/.github/workflows/build-templates.yml
```diff
@@ -0,0 +1,45 @@
+name: Build and Upload Templates
+on:
+  schedule:
+    - cron: "0 4 * * *" # Nightly at 4 AM
+  workflow_dispatch:
```
runleveldev: Also run on push so we can test it in the branch. Might want to add some logic that only actually runs it if any of the relevant files are changed. i.e. any thing under this directory. That way pushes that only affect files in say create-a-container don't trigger a rebuild.

container-creation/intern-phxdc-pve1/packer/api/proxmox_utils.py
```diff
@@ -0,0 +1,49 @@
...
+            data=data,
+            verify=False
+        )
+        r.raise_for_status()
+        return r.json()
+
+def choose_default_storage(storages):
+    """Pick the first local-type storage, or fallback to 'local'."""
+    for s in storages:
+        if s == "local" or "local" in s.lower():
```
runleveldev: Just a question: should `local` be our default? Since `local` is on `/` traditionally it might not be the best place for these. Maybe we detect the first non-`local` storage instead as the default? IDK.

container-creation/intern-phxdc-pve1/packer/provisioners/ansible/site.yml
```diff
@@ -0,0 +1,56 @@
...
+        content: |
+          This will be replaced by logic to deploy container information.
+          This container image was built automatically via Packer and Ansible.
+
+    - name: Configure system DNS (common)
+      ansible.builtin.copy:
+        dest: /etc/resolv.conf
+        content: |
+          nameserver 1.1.1.1
+          nameserver 8.8.8.8
```
runleveldev: We definitely shouldn't be doing this. DNS needs to come from DHCP for this to work. The Proxmox LXC template should havve all that setup correctly, so let's avoid touching anything network related unless we have to in the provisioning.

container-creation/intern-phxdc-pve1/packer/provisioners/ansible/site.yml
```diff
@@ -0,0 +1,56 @@
...
+      when: ansible_os_family in ['Debian', 'RedHat']
+
+    - name: Copy pown.sh (disabled until first boot)
+      ansible.builtin.copy:
+        src: files/pown.sh
+        dest: /usr/local/bin/pown.sh
```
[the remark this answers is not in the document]
runleveldev: Also we're missing `files/pown.sh` so this wouldn't work as-is anyways.

container-creation/intern-phxdc-pve1/packer/provisioners/ansible/site.yml
```diff
@@ -0,0 +1,56 @@
...
+        mode: '0755'
+
+    - name: Copy Wazuh placeholder registration script
+      ansible.builtin.copy:
+        dest: /etc/wazuh-agent/init_disabled
+        content: |
+          #!/bin/bash
+          # Registration disabled in build context; runs on first boot only.
```
runleveldev: Does this need to be a specific agent registration script? I thought we had that covered in https://github.com/mieweb/opensource-server/blob/5d553cc5c875b622b40e0268bdd6af61c29c6e89/Wazuh/register-agent.sh (FYI, I accidentally deleted this file when documenting the container creation process, I'll add it back to the repo soon). We should copy lines 18-24 to install the Wazuh agent and register it to start at the first boot of the container. Since this build happens offline, it shouldn't try to register with the manager until then.

runleveldev (changes requested):
Also these files are all in the `container-creation/intern-phxdc-pve1` directory if you didn't notice. Let's get them their own `packer` or `templates` or `images` subdir and merge the `.github` directories.

container-creation/intern-phxdc-pve1/packer/provisioners/ansible/site.yml
cmyers-mieweb: Removed from the playbook in next commit

container-creation/intern-phxdc-pve1/packer/provisioners/ansible/site.yml
cmyers-mieweb: I think the goal of this was to brainstorm some sort of way to migrate from provisioning from bash scripts, and moved toward a more IaC solution and to only run the provisioning of the agent during the container creation process and not during the nightly template creation to prevent making an unnecesary agent for the template container

runleveldev (approved):
LGTM

.github/workflows/build-templates.yml
```diff
@@ -0,0 +1,50 @@
...
+jobs:
+
+  build:
+    runs-on: ubuntu-latest
+    env:
+      PROXMOX_API_URL: ${{ secrets.PROXMOX_API_URL }}
+      PROXMOX_TOKEN_ID: ${{ secrets.PROXMOX_TOKEN_ID }}
+      PROXMOX_TOKEN_SECRET: ${{ secrets.PROXMOX_TOKEN_SECRET }}
+      # Define the template version ONCE for the whole job
+      TEMPLATE_VERSION: ${{ format('{0:yyyyMMdd}', github.run_started_at) }}
```
runleveldev: The template is not valid. .github/workflows/build-templates.yml (Line: 20, Col: 25): The format specifiers 'yyyyMMdd' are not valid for objects of type 'Null'

runleveldev (changes requested):

.github/workflows/build-templates.yml
cmyers-mieweb: Modified TEMPLATE_VERSION to use shell date instead
`      - name: Set TEMPLATE_VERSION

        run: echo "TEMPLATE_VERSION=$(date +%Y%m%d)" >> $GITHUB_ENV`

runleveldev (approved):
LGTM

PR #90, opened 2025-10-28
refactor: update DHCP range and options in dnsmasq configuration

runleveldev:
This pull request updates the DHCP configuration in `dnsmasq.conf` to support multiple network ranges and assign router options based on network tags. The main changes are the addition of a new DHCP range for a second subnet and the use of tags to specify router options per subnet.

**DHCP configuration enhancements:**

* Added a new DHCP range for the `10.16.0.0/16` subnet with the tag `net16`, alongside the existing `10.15.0.0/16` range now tagged as `net15`.
* Updated DHCP router options to use tags, so that devices in `net15` receive `10.15.0.1` as their router and devices in `net16` receive `10.16.0.1`.

### DIFF
--- a/dnsmasq-service/dnsmasq.conf
+++ b/dnsmasq-service/dnsmasq.conf
@@ -7,7 +7,8 @@ listen-address=127.0.0.1,10.15.0.3
 server=8.8.8.8
 
 # DHCP range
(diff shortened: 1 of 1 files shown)


cmyers-mieweb (approved):
LGTM

Issue #82, opened 2025-10-28
EW-AI-Server SSH Authentication issues present


cmyers-mieweb:
Doug pointed out on 10/27/25 that users are experiencing issues SSH'ing into their containers created on the fort wayne AI server. This looks to be an issue tied with multiple components
* /etc/sssd/sssd.conf is not populating on container creation
* /etc/sssd is set to the wrong permissions to start the service
* sssd is not getting properly installed on the system

After recreating the issue I found these to be the culprits and after performing some test fixes was able to get SSH authentication back working.

<img width="951" height="255" alt="Image" src="https://github.com/user-attachments/assets/7b8ad60c-37b1-483d-b00f-79c9ef71c117" />

PR #96, opened 2025-10-28
Update Fort Wayne container creation and iptables rules

cmyers-mieweb:
https://github.com/mieweb/opensource-server/issues/82

Changed Fort Wayne template CTID to 20113 and adjusted container ID allocation in create-container-new.sh. Updated register-container.sh to use interface-specific iptables rules (wg0 for Fort Wayne, vmbr0 otherwise) for PREROUTING and POSTROUTING, improving network configuration flexibility.

This change also indirectly includes an updates container template on ew-ai that fixes sssd issues related to authentication.

### DIFF
--- a/container-creation/intern-phxdc-pve1/var-lib-vz-snippets/register-container.sh
+++ b/container-creation/intern-phxdc-pve1/var-lib-vz-snippets/register-container.sh
@@ -109,7 +109,7 @@ else
     fi
 
         # SSH PREROUTING rule
-        iptables -t nat -A PREROUTING -i "$IPTABLES_IFACE" -p tcp --dport "$ssh_port" -j DNAT --to-destination "$container_ip:22"
+        iptables -t nat -A PREROUTING -i vmbr0 -p tcp --dport "$ssh_port" -j DNAT --to-destination "$container_ip:22"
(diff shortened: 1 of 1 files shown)


container-creation/intern-phxdc-pve1/var-lib-vz-snippets/register-container.sh
```diff
@@ -101,12 +108,12 @@ else
         exit 2
     fi
 
-    # Add PREROUTING rule
-    iptables -t nat -A PREROUTING -i vmbr0 -p tcp --dport "$ssh_port" -j DNAT --to-destination "$container_ip:22"
+        # SSH PREROUTING rule
+        iptables -t nat -A PREROUTING -i "$IPTABLES_IFACE" -p tcp --dport "$ssh_port" -j DNAT --to-destination "$container_ip:22"
```
runleveldev: This should stay vmbr0 regardless of the location. PREROUTING applies to incomming connections which are always established on vmbr0, we only need to update POSTROUTING which applies to outgoing connections which is what changes based on location.

runleveldev (approved):
LGTM

PR #98, opened 2025-10-29
add default SSL configuration for Nginx reverse proxy

runleveldev:
This pull request adds a new Nginx SSL configuration file to enable secure connections with modern TLS, HTTP/2, and HTTP/3 support, as well as improved security headers and certificate management.

**New SSL server configuration:**

* Introduced `default-ssl.conf` to configure Nginx to listen on port 443 with SSL, HTTP/2, and HTTP/3 (QUIC) enabled.
* Configured SSL certificates and modern TLS protocols/ciphers for strong encryption.
* Enabled SSL session optimization and OCSP stapling for improved performance and certificate validation.
* Added security headers such as HSTS, X-Frame-Options, X-Content-Type-Options, and X-XSS-Protection for better browser security.
* Configured the server to return a 404 error for unhandled hostnames, preventing accidental exposure of the default server.

### DIFF
--- a/nginx-reverse-proxy/conf.d/default-ssl.conf
+++ b/nginx-reverse-proxy/conf.d/default-ssl.conf
@@ -1,0 +1,41 @@
+server {
+  listen 443 ssl;
+  listen [::]:443 ssl;
+  listen 443 quic;
+  listen [::]:443 quic;
+  http2 on;
+  http3 on;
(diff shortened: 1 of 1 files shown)


cmyers-mieweb (approved):
Changes approved and applied in prod.

Issue #115, opened 2025-11-14
New Confirmation Prompts break/hang Container Creation.


cmyers-mieweb:
I Identified that the new confirmation prompts presented when running pown.sh aren't compatible with automated builds, such as seen on the opensource cluster. There needs to be a `-y` option to bypass interactive configurations, or something similar in nature.

<img width="421" height="176" alt="Image" src="https://github.com/user-attachments/assets/1e02d14d-7935-43ff-8b78-c6f2b9a2ad63" />

<img width="969" height="132" alt="Image" src="https://github.com/user-attachments/assets/9e2e400d-d09e-4d0b-9a6f-4ad5418e56f7" />

PR #117, opened 2025-11-14
Import .env file into container during LDAP setup

cmyers-mieweb:
https://github.com/mieweb/opensource-server/issues/115

Added logic to configureLDAP.sh to import a .env file (pown.sh.env) from the host into the container at /etc/pown.sh.env. This ensures environment variables needed for LDAP/SSSD configuration are available inside the container while we wait for a non interactive mode.


runleveldev (approved):
LGTM

PR #136, opened 2025-11-21
118 bootstrap procedure

runleveldev:
Required features:

- [x] 'Site' model (id PK, name, internalDomain, subnet, gateway, dnsForwarders). FRE configures (1, 'default', <selected domain>, <confirms detected subnet>, <confirm detected gateway>, <confirm detected dns forwarders>)
- [x] 'Group' model (gidNumber PK, cn, isAdmin), FRE configures [(GID_MIN, 'sysadmins', True), (GID_MIN + 1, 'ldapusers', False)]
- [x] 'User' model, fields based on LDAP schemas (uidNumber PK, uid, gidNumber, homeDirectory, loginShell, cn, sn, givenName, mail, sshPublicKey, userPassword, status), FRE configures the first user using UID_MIN.
- [x] 'UserGroups' association model (uidNumber, gidNumber). Users get 'ldapusers' as primary group by default. 'sysadmins' denotes cluster superusers. The 'ldapusers' group will get SUDO on containers, but only 'sysadmins' will on Nodes and core infra containers
- [x] 'ExternalDomain' model (id, name, cloudflareApiKey)
- [x] Alter 'Service' model add externalDomainId to http services
- [ ] ~Alter 'Container' model replace username with userId~
- [x] Create dnsmasq.conf template
- [x] Some sort of acme configuration, `lego`
- [x] Node import
- [x] Container import
- [x] Implement API-backed container creation


cmyers-mieweb:
Did we want to also include additional rate limiting for logins and registers inside `/create-a-container/views/server.js`?
Something similar to adding this `const`
```
  const authLimiter = RateLimit({
    windowMs: 15 * 60 * 1000, // 15 minutes
    max: 10, // max 10 attempts per IP per window
    standardHeaders: true,
    legacyHeaders: false
  });
```

and then applying it to lines 116 and 117 `app.use('/login', loginRouter);` and `app.use('/register', registerRouter);`
```
  app.use('/login', authLimiter, loginRouter);
  app.use('/register', authLimiter, registerRouter);
 ```

PR #157, opened 2025-12-30
Add Templates feature for container creation

runleveldev:
- Create Template model with displayName, proxmoxTemplateName, and siteId
- Add migration to create Templates table
- Create templates router with full CRUD operations (admin only)
- Add getAvailableProxmoxTemplates helper function to query nodes
- Create templates index and form views
- Update header to show Templates link for admins
- Update container creation to use Template model instead of querying Proxmox directly
- Mount templates router in sites router


runleveldev:
Not needed. #189

PR #160, opened 2026-01-02
switch to lxc-backed templates

runleveldev:
* Fixes error handling on create container route
* Clones existing LXC as template instead of using prebuilt tar file templates


Bobzemob (approved):
LGTM

Issue #172, opened 2026-01-28
Integrate API changes to integrate Launchpad builds


cmyers-mieweb:
Launchpad will now use an API key for action deployments. Currently the /containers route cannot GET and a new API route will need to be created so that a simple API can exist for the action to query. 
```
Run mieweb/launchpad@cmyers_issue3
  with:
    api_key: ***
    api_url: ***
    template_name: ubuntu-22.04-standard
    install_command: apt-get update && apt-get install -y golang
    build_command: go build -o whoami .
    start_command: ./whoami --port 80
    runtime_language: go
    container_env_vars: {"WHOAMI_NAME": "LaunchPad-Test"}
Run if [[ "$GITHUB_EVENT_NAME" != "push" ]] || [[ "$GITHUB_EVENT_CREATED" == "false" ]]; then
Run if [[ "$GITHUB_JOB" == *"setup"* || "$GITHUB_JOB" == *"runner"* ]]; then
Run if [[ "$GITHUB_EVENT_NAME" == "delete" ]]; then
Using current branch name: master
Run # 1. Generate Container Name
Managing Container: cmyers-mieweb-whoami-master
Querying API: ***/containers?hostname=cmyers-mieweb-whoami-master
--- RAW API RESPONSE ---
* Host manager.os.mieweb.org:443 was resolved.
* IPv6: (none)
* IPv4: 184.175.182.248
*   Trying 184.175.182.248:443...
* Connected to manager.os.mieweb.org (184.175.182.248) port 443
* ALPN: curl offers h2,http/1.1
} [5 bytes data]
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
} [512 bytes data]
*  CAfile: /etc/ssl/certs/ca-certificates.crt
*  CApath: /etc/ssl/certs
{ [5 bytes data]
* TLSv1.3 (IN), TLS handshake, Server hello (2):
{ [122 bytes data]
* TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
{ [41 bytes data]
* TLSv1.3 (IN), TLS handshake, Certificate (11):
{ [2944 bytes data]
* TLSv1.3 (IN), TLS handshake, CERT verify (15):
{ [78 bytes data]
* TLSv1.3 (IN), TLS handshake, Finished (20):
{ [52 bytes data]
* TLSv1.3 (OUT), TLS change cipher, Change cipher spec (1):
} [1 bytes data]
[...]

PR #173, opened 2026-01-28
Add API endpoints for container management through Launchpad

cmyers-mieweb:
Issue: https://github.com/mieweb/opensource-server/issues/172

Introduces a new router providing RESTful API endpoints for managing containers, including create, read, update, and delete operations with API key authentication. The new endpoints are mounted at the top level in server.js to support automation and API clients.

Files changed:

api_containers.js : new API router (GET/POST/PUT/DELETE /containers) with Bearer API key auth.
server.js:1 : mounts the new router at top-level so /containers is available.

### DIFF
--- a/create-a-container/middlewares/index.js
+++ b/create-a-container/middlewares/index.js
@@ -14,18 +14,14 @@ async function requireAuth(req, res, next) {
   
   // Try API key authentication
   const authHeader = req.get('Authorization');
-  console.log(`[AUTH DEBUG] Authorization header: "${authHeader || 'NONE'}" (length: ${authHeader?.length || 0})`);
-  
   if (authHeader && authHeader.startsWith('Bearer ')) {
     const apiKey = authHeader.substring(7);
-    console.log(`[AUTH DEBUG] Extracted API key, length: ${apiKey.length}, first 8 chars: ${apiKey.substring(0, 8)}`);
     
-    if (apiKey && apiKey.length > 0) {
+    if (apiKey) {
       const { ApiKey, User } = require('../models');
       const { extractKeyPrefix } = require('../utils/apikey');
       
       const keyPrefix = extractKeyPrefix(apiKey);
-      console.log(`[AUTH DEBUG] Key prefix: ${keyPrefix}`);
       
       const apiKeys = await ApiKey.findAll({
         where: { keyPrefix },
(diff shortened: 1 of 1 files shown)


create-a-container/routers/containers.js
```diff
@@ -65,8 +65,43 @@ router.get('/new', requireAuth, async (req, res) => {
...
+function isApiRequest(req) {
+  const accept = (req.get('accept') || '').toLowerCase();
+  return accept.includes('application/json') || accept.includes('application/vnd.api+json');
+}
+
 // GET /sites/:siteId/containers - List all containers for the logged-in user in this site
-router.get('/', requireAuth, async (req, res) => {
+router.get('/', async (req, res) => {
+  // If called by API clients using Bearer token, return JSON instead of HTML
+  if (isApiRequest(req)) {
```
runleveldev: Instead of duplicating work in this `if` branch, replace the other `return`s in this function depending on whether this is an API request. As an example, we currently have `return res.redirect('/sites');` on line 75, replace that with:

```
if (isApiRequest(req)) {
  return res.status(404).json({error: "Site not found"});
} else {
  return res.redirect('/sites');
}
```

And, on line 158 we have `return res.render(...` which could be replaced with

```
if (isApiRequest(req)) {
  return res.json({containers: rows});
} else {
  return res.render(...);
}
```

create-a-container/routers/containers.js
```diff
@@ -65,8 +65,43 @@ router.get('/new', requireAuth, async (req, res) => {
   });
 });
 
+// Helper to detect API bearer requests
+function isApiRequest(req) {
```
runleveldev: As is, this should be moved to a util function and imported here instead of living in this file.

create-a-container/routers/containers.js
```diff
@@ -513,9 +593,21 @@ router.put('/:id', requireAuth, async (req, res) => {
 });
 
 // DELETE /sites/:siteId/containers/:id - Delete a container
-router.delete('/:id', requireAuth, async (req, res) => {
+router.delete('/:id', async (req, res) => {
   const siteId = parseInt(req.params.siteId, 10);
   const containerId = parseInt(req.params.id, 10);
+  // If API request, perform lightweight delete and return JSON/204
+  if (isApiRequest(req)) {
```
runleveldev: See previous comments about return locations.

create-a-container/routers/containers.js
```diff
@@ -513,9 +593,21 @@ router.put('/:id', requireAuth, async (req, res) => {
 });
 
 // DELETE /sites/:siteId/containers/:id - Delete a container
-router.delete('/:id', requireAuth, async (req, res) => {
+router.delete('/:id', async (req, res) => {
```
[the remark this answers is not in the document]
cmyers-mieweb: I readded this to the associated router.delete methods removed

create-a-container/routers/containers.js
cmyers-mieweb:
```
// GET /sites/:siteId/containers/new - List available templates via API or HTML form
router.get('/new', requireAuth, async (req, res) => {
  const siteId = parseInt(req.params.siteId, 10);
  const isApi = isApiRequest(req); 

  // verify site exists
  const site = await Site.findByPk(siteId);
  if (!site) {
    if (isApi) return res.status(404).json({ error: 'Site not found' });
    await req.flash('error', 'Site not found');
    return res.redirect('/sites');
  }
  
  ```
  ```
  // GET /sites/:siteId/containers/:id/edit
router.get('/:id/edit', requireAuth, async (req, res) => {
  const siteId = parseInt(req.params.siteId, 10);
  const containerId = parseInt(req.params.id, 10);
  
  const site = await Site.findByPk(siteId);
  if (!site) {
    await req.flash('error', 'Site not found');
    return res.redirect('/sites');
  }

```

create-a-container/server.js
```diff
@@ -103,6 +103,8 @@ async function main() {
   const groupsRouter = require('./routers/groups');
   const sitesRouter = require('./routers/sites'); // Includes nested nodes and containers routers
   const jobsRouter = require('./routers/jobs');
+  // expose API endpoints before HTML routes so they respond at top-level
+  app.use('/', apiContainersRouter);
```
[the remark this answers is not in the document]
cmyers-mieweb: Removed

PR #183, opened 2026-02-06
fix: implement proper registry auth flow

runleveldev:
Fixes container template pulling from ghcr.io by implementing the v2 registry auth spec.

https://docs.docker.com/reference/api/registry/auth/


cmyers-mieweb (approved):
Biggest concern was the inclusion of the API key authentication changes in another recent PR, but after testing this can be merged fine.

PR #252, opened 2026-03-20
feat: Add quick login button for dev mode

horner:
Add one-click quick login for local development - shows split-button dropdown with Admin/Standard User options when no .env exists. First user becomes admin automatically. Includes Playwright tests.


runleveldev:
Closing in preference of #304

PR #274, opened 2026-04-02
Add LDAP-based git identity and SSSD config for immediate git config setup

### LINKED ISSUE #256: Containers should have git pre-configured
I have to run 

`git config user.name "Doug Horner" && git config user.email "horner@mieweb.com"`

everytime in the container.  I want to make it so we you log into the container for the first time, the identity in the manager propagates as I log in from container to container.  I would be best if it works with any template and for users regardless if they created the container or not.

cmyers-mieweb:
Resolves: https://github.com/mieweb/opensource-server/issues/256

Installs ldap-utils and adds LDAP client/config and a profile script to auto-configure git user.name/email from LDAP on first interactive login. 

Copies ldap.conf to /etc/ldap, adds /etc/profile.d/git-identity.sh which uses ldapsearch and NSS (sssd) gecos to set global git config, and adjusts sssd.conf to map cn -> gecos (ldap_user_gecos = cn). 

Also updates Dockerfile to install ldap-utils and include the new files.

This should allow for any user logging into any container to have `git config` preset and ready to go. This should work on any template derived from the base image.


images/base/git-identity.sh
```diff
@@ -0,0 +1,40 @@
...
+
+_GIT_SETUP_USER="${USER:-$(id -un 2>/dev/null)}"
+[ -z "$_GIT_SETUP_USER" ] && return
+
+# Full name from NSS (SSSD maps LDAP cn -> gecos via ldap_user_gecos = cn)
+_GIT_SETUP_NAME=$(getent passwd "$_GIT_SETUP_USER" 2>/dev/null | cut -d: -f5)
+
+# Email from LDAP anonymous query
+_GIT_SETUP_LDAP_HOST="${LDAP_URI:-ldaps://ldap1:636}"
+_GIT_SETUP_LDAP_BASE="${LDAP_BASE_DN:-dc=docker,dc=internal}"
```
runleveldev: BaseDN is a problem here. We're allowing SSSD to do baseDN autodiscovery via rootDSE namingContext attribute. To match the SSSD config's baseDN, you would need to query the rootDSE, use namingContext if there's only one, otherwise use defaultNamingContext if theres multiple namingContexts otherwise fail (because SSSD would have failed too).

images/base/git-identity.sh
```diff
@@ -0,0 +1,40 @@
...
+
+# Only run for interactive shells
+[[ $- != *i* ]] && return
+
+# Only if git is available
+command -v git >/dev/null 2>&1 || return
+command -v ldapsearch >/dev/null 2>&1 || return
+
+# Skip if already configured — user-set values always take precedence
+[ -n "$(git config --global user.email 2>/dev/null)" ] && return
```
runleveldev: We should probably only skip if user.email AND user.name are set globally.

images/base/git-identity.sh
```diff
@@ -0,0 +1,40 @@
...
+
+_GIT_SETUP_USER="${USER:-$(id -un 2>/dev/null)}"
+[ -z "$_GIT_SETUP_USER" ] && return
```
runleveldev, line 21: Can we bailout here as well if user is root? Just to avoid the unnessecary LDAP lookup?

images/base/sssd.conf
```diff
@@ -7,5 +7,9 @@ auth_provider = ldap
 ldap_uri = ldaps://ldap1:636, ldaps://ldap2:636
 ldap_tls_reqcert = allow
 
+# Map LDAP cn attribute to the NSS gecos field so that tools like getent,
+# finger, and the git-identity profile script can read the user's full name.
+ldap_user_gecos = cn
```
runleveldev, line 12: The default value of this setting `gecos` works for our deployment. I don't want to complicate the sssd config more than nessecary.

runleveldev (changes requested):

images/base/git-identity.sh
```diff
@@ -0,0 +1,40 @@
...
+_GIT_SETUP_LDAP_BASE="${LDAP_BASE_DN:-dc=docker,dc=internal}"
+_GIT_SETUP_EMAIL=$(ldapsearch -x \
+    -H "$_GIT_SETUP_LDAP_HOST" \
+    -b "$_GIT_SETUP_LDAP_BASE" \
+    "(uid=${_GIT_SETUP_USER})" mail 2>/dev/null \
+    | awk '/^mail:/{print $2; exit}')
+
+if [ -n "$_GIT_SETUP_NAME" ] && [ -n "$_GIT_SETUP_EMAIL" ]; then
+    git config --global user.name  "$_GIT_SETUP_NAME"
+    git config --global user.email "$_GIT_SETUP_EMAIL"
```
cmyers-mieweb: Separate these, email isn't going to be required out of the box. Git can handle this if it isn't present.

images/base/git-identity.sh
cmyers-mieweb: Created some conditionals to resolve this issue. 
```
# Resolve baseDN the same way SSSD does: rootDSE namingContexts autodiscovery.
# Use LDAP_BASE_DN if explicitly set; otherwise query rootDSE.
# - Single namingContexts entry  -> use it directly
# - Multiple namingContexts      -> use defaultNamingContext
# - Neither resolvable           -> abort
if [ -n "${LDAP_BASE_DN:-}" ]; then
    _GIT_SETUP_LDAP_BASE="$LDAP_BASE_DN"
else
    _GIT_SETUP_ROOTDSE=$(ldapsearch -x -H "$_GIT_SETUP_LDAP_HOST" -b "" -s base namingContexts defaultNamingContext 2>/dev/null)
    _GIT_SETUP_NC_COUNT=$(echo "$_GIT_SETUP_ROOTDSE" | grep -c '^namingContexts:')
    if [ "$_GIT_SETUP_NC_COUNT" -eq 1 ]; then
        _GIT_SETUP_LDAP_BASE=$(echo "$_GIT_SETUP_ROOTDSE" | awk '/^namingContexts:/{print $2; exit}')
    elif [ "$_GIT_SETUP_NC_COUNT" -gt 1 ]; then
        _GIT_SETUP_LDAP_BASE=$(echo "$_GIT_SETUP_ROOTDSE" | awk '/^defaultNamingContext:/{print $2; exit}')
    fi
    unset _GIT_SETUP_ROOTDSE _GIT_SETUP_NC_COUNT
fi
[ -z "${_GIT_SETUP_LDAP_BASE:-}" ] && return
```

images/base/sssd.conf
cmyers-mieweb: I went ahead and removed these lines

images/base/git-identity.sh
cmyers-mieweb:
```
_GIT_SETUP_USER="${USER:-$(id -un 2>/dev/null)}"
[ -z "$_GIT_SETUP_USER" ] && return
[ "$_GIT_SETUP_USER" = "root" ] && return
```

images/base/git-identity.sh
cmyers-mieweb: changed to 
`[ -n "$(git config --global user.email 2>/dev/null)" ] && [ -n "$(git config --global user.name 2>/dev/null)" ] && return`

runleveldev (approved):

PR #301, opened 2026-05-11
create containers on nodes with least current containers

### LINKED ISSUE #299: Distribute containers across multiple nodes
On https://github.com/mieweb/opensource-server/blob/ee76e99b47b6ee6f4bcbea173a7c3cc9ebff90c3/create-a-container/routers/containers.js#L354 we get one node that matches the condition, instead we should be finding the node that matches the condition which has the least amount of containers currently.

runleveldev:
You can see the node placement logic at work here. Manager and demo already existed on pve1, so pve2 got filled before pve1 started to be used again.

image

# Copilot Summary

This pull request improves the logic for selecting a node when creating a new container, aiming to better balance the load across available nodes. Instead of picking any available node, the system now chooses the node with the fewest existing containers.

**Load balancing improvements:**

* Updated the node selection query in `containers.js` to join with the `Container` model, group by node, and order by the number of containers, ensuring the least-loaded node is chosen for new containers.

### DIFF
--- a/create-a-container/routers/containers.js
+++ b/create-a-container/routers/containers.js
@@ -351,7 +351,26 @@ router.post('/', async (req, res) => {
       nodeWhere.nvidiaAvailable = true;
     }
 
-    const node = await Node.findOne({ where: nodeWhere });
+    // Select the node with the fewest existing containers to balance load.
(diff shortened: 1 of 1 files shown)


cmyers-mieweb (approved):
Looks good, lets get it merged

Issue #305, opened 2026-05-13
Feature: Add 'Email all' button to Users tab to send email to all users


cmyers-mieweb:
**Feature Request**

In the 'Users' tab, add a button next to the existing 'Invite' and 'New User' buttons labeled "Email all". When clicked, this button should:

1. Open a modal or popup with:
    - A required Subject field
    - A required Message field
2. Upon completion and submission:
    - An email is sent from the opensource-server to all users in the current users list.

**Why?**
This will allow admins to quickly send an email to all registered users directly from the user interface, improving communication and workflow.

**Acceptance Criteria:**
- The 'Email all' button appears next to 'Invite' and 'New User' in the Users tab.
- Modal includes Subject and Message fields (both required).
- Submitting form sends an email to all users through the opensource-server's email system.
- Proper error handling for failed emails.

**Additional Notes:**
- Please consider confirmation before sending the email.
- Ensure only authorized users (admins) can access this functionality.
- UI/UX should align with current tab style.

runleveldev:
Completed in #306

PR #306, opened 2026-05-13
Add 'Email All' feature for users

cmyers-mieweb:
issue: https://github.com/mieweb/opensource-server/issues/305
Add UI, route, and email utility to send a message to all users with an email address. Changes include:

- routers/users.js: import sendBulkEmail and add POST /users/email-all which validates subject/message, checks SMTP configuration, collects unique non-empty user emails, calls sendBulkEmail, and sets flash messages for success/partial/failure cases with error logging.
- utils/email.js: add sendBulkEmail (sends one message per recipient using configured noreply address), escapeHtml helper to sanitize HTML body, and export sendBulkEmail.
- views/users/index.ejs: add "Email All" button and modal form to compose subject and message, showing recipient count and confirming send.

Includes basic input validation, SMTP existence check, per-recipient send reporting, and HTML-escaping of message content.
image

### DIFF
--- a/create-a-container/utils/email.js
+++ b/create-a-container/utils/email.js
@@ -204,7 +204,7 @@ async function sendBulkEmail(recipients, subject, message) {
   await transporter.sendMail({
     from,
     to: from,
-    bcc: recipients.join('; '),
+    bcc: recipients.join(', '),
     subject,
     text: message,
     html
--- a/create-a-container/utils/email.js
+++ b/create-a-container/utils/email.js
@@ -154,8 +154,68 @@ Medical Informatics Engineering`,
   await transporter.sendMail(mailOptions);
 }
 
+/**
+ * Escape HTML special characters to prevent injection in email HTML body
+ */
+function escapeHtml(str) {
+  return String(str)
+    .replace(/&/g, '&amp;')
+    .replace(/</g, '&lt;')
+    .replace(/>/g, '&gt;')
+    .replace(/"/g, '&quot;')
+    .replace(/'/g, '&#39;');
+}
+
+/**
+ * Send a bulk email to multiple recipients (one message per recipient).
+ * Returns { sent, failed } where failed is an array of { to, error }.
+ *
+ * @param {string[]} recipients - List of recipient email addresses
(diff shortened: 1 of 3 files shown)


create-a-container/utils/email.js
```diff
@@ -154,8 +154,68 @@ Medical Informatics Engineering`,
...
+ * Send a bulk email to multiple recipients (one message per recipient).
+ * Returns { sent, failed } where failed is an array of { to, error }.
+ *
+ * @param {string[]} recipients - List of recipient email addresses
+ * @param {string} subject - Email subject
+ * @param {string} message - Plain-text message body
+ */
+async function sendBulkEmail(recipients, subject, message) {
+  const settings = await Setting.getMultiple(['smtp_noreply_address']);
+  const from = settings.smtp_noreply_address || 'noreply@localhost';
```
runleveldev: I don't like this fallback. If smtp_noreply_address is not set this should throw up instead.

create-a-container/utils/email.js
```diff
@@ -154,8 +154,68 @@ Medical Informatics Engineering`,
...
+        to,
+        subject,
+        text: message,
+        html
+      });
+      sent.push(to);
+    } catch (error) {
+      failed.push({ to, error: error.message });
+    }
+  }
```
runleveldev: Can we do a single email with BCC instead of looping?

runleveldev (changes requested):

create-a-container/utils/email.js
cmyers-mieweb: consolidated loop into one action
```
  await transporter.sendMail({
    from,
    to: from,
    bcc: recipients.join(', '),
    subject,
    text: message,
    html
  });
```

create-a-container/utils/email.js
cmyers-mieweb: modified each occurrence to throw and error instead
```
  const from = settings.smtp_noreply_address;
  if (!from || from.trim() === '') {
    throw new Error('SMTP no-reply address is not configured');
  }
```

runleveldev (approved):

Issue #317, opened 2026-05-18
Multiple containers created through Launchpad API


runleveldev:
<img width="506" height="249" alt="Image" src="https://github.com/user-attachments/assets/9b2daf86-a7f0-4633-84f7-7652d1965145" />

PR #318, opened 2026-05-18
Prevent orphaned containers and duplicate hosts

cmyers-mieweb:
Issue: https://github.com/mieweb/opensource-server/issues/317

- Add safeguards to container creation and deletion to avoid orphaned Proxmox containers and duplicate hostnames.
- Import `Job` model and check for cancelled creation jobs during provisioning; if cancelled, abort creation.
- After provisioning but before applying config, verify the DB record still exists and if it was deleted, destroy the orphaned Proxmox container and abort.
- In the API v1 containers router, enforce a unique hostname-per-site check (with transaction lock) and return `409` for duplicate hostnames.
- Cancel any pending/running creation job when a container is deleted.
- Resolve merge conflicts with `main` and preserve these safeguards in the updated router layout.

**Changes affect**

1. `create-a-container/bin/create-container.js`
2. `create-a-container/routers/api/v1/containers.js`


cmyers-mieweb:
@copilot resolve the merge conflicts in this pull request

create-a-container/routers/api/v1/containers.js
```diff
@@ -413,6 +413,18 @@ router.post(
...
+        transaction: t,
+        lock: t.LOCK.UPDATE,
+      });
+      if (existingContainer) {
+        throw new ApiError(
+          409,
+          'duplicate_hostname',
+          `Container with hostname "${hostname}" already exists in this site`,
+        );
+      }
```
runleveldev, line 427: Container should have a unique condition in the database on siteId + hostname already. The create statement on line 466 below should throw in this case. This preflight check adds an extra DB roundtrip for something that should be caught there.

runleveldev (changes requested):

create-a-container/bin/create-container.js
```diff
@@ -230,6 +230,15 @@ async function main() {
...
     const vmid = await client.nextId();
     console.log(`Allocated VMID: ${vmid}`);
+
+    // Check if our job was cancelled (e.g., container was deleted while we were starting)
+    if (container.creationJobId) {
+      const job = await Job.findByPk(container.creationJobId);
+      if (job && job.status === 'cancelled') {
+        console.error('Job was cancelled — aborting container creation.');
+        process.exit(1);
+      }
```
runleveldev, line 240: Doing this would probably completely rewrite this PR but could we instead have job-runner watch for cancelled jobs and send SIGTERM to them? Then this job would have a SIGTERM handler to rollback it's changes before exitting.

runleveldev (changes requested):

Issue #313, opened 2026-05-14
Clean up manager logs


runleveldev:
The manager logs all accesses to stdout, causing noisy logs in journalctl in production. Access logs should be moved to a dedicate file (ie. /var/log/opensource-server/access.log) and a logrotate config should be provided. For consistency, a log file may be created for errors as well (ie. /var/log/opensource-server/error.log). Both logs should be made available as viewable to administrators via the Web UI.

Issue #311, opened 2026-05-14
CSS issue


runleveldev:
Certain pages' UI elements are shifted off screen at thin resolutions.

Issue #310, opened 2026-05-14
Surface better errors on database constraint failures


runleveldev:
User reported container could not be created due to "validation error". This should instead provide a clear error message to the user (regardless of the specific constraint being violated) instead of the generic error message.

Issue #309, opened 2026-05-14
[Bug]: MIEAuth issue


horner:
### Current URL

/users/2043/edit

### Username

horner

### Version

v2026.05.11-1 (2b93f27)

### Steps to reproduce

https://mieauth-prod.os.mieweb.org/

<img width="600" alt="Image" src="https://github.com/user-attachments/assets/8f245137-c283-45bb-b376-1fb620ae4c72" />

<img width="600" alt="Image" src="https://github.com/user-attachments/assets/56fc5358-d674-49e9-b294-bc3ec1fa516e" />

### Expected behavior

How do I approve

### Actual behavior

Cant find how to approve

PR #319, opened 2026-05-19
React SPA cutover: replace EJS views with /api/v1 + React 19 client

wreiske:
## Summary

This branch cuts the `create-a-container` app over from server-rendered EJS views to a React 19 SPA backed by a new versioned JSON API at `/api/v1`. Every legacy resource page has a React equivalent and the EJS view layer has been removed.

## Highlights

**Backend — `/api/v1` JSON API**
- New routers under `create-a-container/routers/api/v1/` for auth, users, groups, sites, nodes, containers, external-domains, jobs, settings, and api-keys
- CSRF, session, and API-key auth wired through new `middlewares/api.js`
- OpenAPI spec at `openapi.v1.yaml`
- Legacy EJS routers and views removed
- A few migrations adjusted to match the new constraints (site-scoped containers, nullable external-domain site, removal of node-name unique index)

**Frontend — React 19 + Vite + Tailwind 4 SPA**
- Scaffolded with `@mieweb/ui`, React Router 7, TanStack Query, react-hook-form + zod
- Pages for every resource: containers, nodes, sites, groups, users, api keys, external domains, jobs, settings, plus full auth flow (login/register/reset)
- Shared `FormPageHeader` / `FormPageLayout` components and `useDocumentTitle` hook for consistent create/edit pages
- Two-panel marketing auth layout with mobile fallback
- Responsive list/table action rows
- Vite dev proxy scoped to `/api/*` so client routes like `/apikeys` aren't intercepted

**Other**
- Legacy EJS screenshots stashed under `.attic/` for reference

## Verification
- `npm run type-check` (client) — clean
- `npm run build` (client) — succeeds
- `node --check` across `server.js`, `routers/api/v1/*.js`, and `middlewares/api.js` — clean

## Commits
- `feat(client): scaffold React 19 + Tailwind 4 + @mieweb/ui SPA`
- `feat(api): add /api/v1 JSON API + CSRF + OpenAPI spec`
- `feat(client): SPA shell + auth flows on /api/v1`
- `feat(client): Phase 4 — feature pages for all resources`
- `feat: cutover from EJS to React SPA`
- `fix(dev): SPA boots end-to-end on sqlite`
- `fix(client): align top bar with sidebar header, polish footer, fix mobile`
- `feat(client): polish auth layout, form scaffolding, and responsive UI`

Screenshot 2026-05-19 at 12 07 09 AM

Screenshot 2026-05-19 at 12 17 02 AM

Screenshot 2026-05-19 at 12 28 24 AM
Screenshot 2026-05-19 at 12 28 31 AM
Screenshot 2026-05-19 at 12 28 35 AM

Screenshot 2026-05-19 at 12 28 00 AM
Screenshot 2026-05-19 at 12 28 10 AM


wreiske:
Merged latest `main` and ported every feature that landed inside the now-deleted EJS code to the new `/api/v1` + React SPA. Branch is now at parity with `main`.

**Ported in `97190e4`:**
- **One-click dev login** (#304) — `POST /api/v1/auth/dev` (404 in production) + Login page buttons gated on `GET /api/v1/health` → `isDev`.
- **Email All** (#306) — `POST /api/v1/users/email-all` using the existing `sendBulkEmail` (BCC + `smtp_noreply_address`) + new `EmailAllModal` wired into the Users list with a live recipient count.
- **MFA Admin link** (#309) — `GET /api/v1/session` now returns `pushNotificationUrl` for admins; Sidebar renders an external link to \`\${pushNotificationUrl}/admin\` when configured.

**Already covered, no port needed:**
- #310 plain-english SQL validation errors — structurally superseded by \`jsonErrorHandler\` returning HTTP 422 with per-field \`fields\` on \`SequelizeValidationError\`/\`SequelizeUniqueConstraintError\`.
- #311 sidebar collapse fix — the new \`@mieweb/ui\` \`Sidebar\` handles responsive collapse natively.
- #313 dedicated access log, Debian env filepath, SMTP no-reply enforcement / BCC bulk batching — auto-merged cleanly into \`server.js\`, \`systemd/\`, and \`utils/email.js\`.

Typecheck + Vite build are clean. \`git diff origin/main..HEAD\` now shows only the expected EJS deletes plus the SPA + JSON API surface.

create-a-container/server.js
```diff
@@ -179,21 +111,16 @@ async function main() {
     customSiteTitle: 'Create-a-Container API',
   }));
```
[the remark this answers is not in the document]
wreiske: Fixed — Swagger UI at `/api` now loads `openapi.v1.yaml`, matching `/api/v1/openapi.*`.

create-a-container/client/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Develop
+
+```bash
+npm install
+npm run dev          # http://localhost:5173 (proxies /api, /login, /logout, /nginx-conf, /dnsmasq to Express)
+```
+
+Express must be running on `http://localhost:3000` (or set `VITE_API_TARGET`).
+
```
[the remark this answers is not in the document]
wreiske: Fixed — README now reflects the actual proxy config (`/api/*` only; other routes are SPA-handled).

create-a-container/client/src/lib/toast.ts
```diff
@@ -0,0 +1,17 @@
...
+  const { error } = useToast();
+  return (err: unknown) => {
+    const e = err as ApiError;
+    if (e && typeof e === 'object' && 'message' in e) {
+      error(e.message || 'Request failed');
+    } else {
+      error('Request failed');
+    }
+  };
+}
```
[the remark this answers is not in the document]
wreiske: Fixed — removed the unused `useApiErrorToast` helper (and the now-empty `lib/toast.ts`).

create-a-container/client/src/app/PlaceholderPage.tsx
```diff
@@ -0,0 +1,12 @@
...
+export function PlaceholderPage({ title }: { title: string }) {
+  const { pathname } = useLocation();
+  return (
+    <section className="space-y-2">
+      <h2 className="text-2xl font-semibold">{title}</h2>
+      <p className="text-sm opacity-70">Route: <code>{pathname}</code></p>
+      <p className="text-sm opacity-70">Phase 1 scaffolding placeholder. Implementation lands in Phase 4.</p>
+    </section>
+  );
+}
```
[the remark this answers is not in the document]
wreiske: Fixed — removed `PlaceholderPage.tsx` (unreferenced).

create-a-container/client/src/pages/users/UsersListPage.tsx
```diff
@@ -0,0 +1,117 @@
...
+                    leftIcon={<Trash2 className="size-4" />}
+                    onClick={() => { if (confirm(`Delete user "${u.uid}"?`)) del.mutate(u.uidNumber); }}
+                    disabled={del.isPending}
+                  >
+      <EmailAllModal
+        open={emailAllOpen}
+        onOpenChange={setEmailAllOpen}
+        recipientCount={data?.filter((u: User) => u.mail?.trim()).length ?? 0}
+      />
+                    Delete
```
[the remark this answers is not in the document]
wreiske: Fixed — moved `EmailAllModal` out of the per-row Delete button and rendered it once at the page root, controlled by `emailAllOpen`.

create-a-container/middlewares/api.js
```diff
@@ -0,0 +1,161 @@
...
+const { doubleCsrf } = require('csrf-csrf');
+
+// --- CSRF (double-submit cookie) ----------------------------------------------------------
+// Token lives in a cookie + must echo in X-CSRF-Token (or _csrf in body).
+// Skipped for Bearer-token API key requests (those are already cryptographically auth'd).
+const csrfSecret = () => process.env.CSRF_SECRET || process.env.SESSION_SECRET || 'dev-csrf-secret-change-me';
+
+// __Host- prefix requires Secure + Path=/ + no Domain; browsers reject it on
+// plain HTTP (including dev), so fall back to a plain cookie name off-prod.
+const isProd = process.env.NODE_ENV === 'production';
```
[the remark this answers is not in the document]
wreiske: Fixed — `csrfSecret()` now throws on startup in `NODE_ENV=production` if neither `CSRF_SECRET` nor `SESSION_SECRET` is set. The dev fallback is preserved off-prod.

create-a-container/client/src/app/Sidebar.tsx
```diff
@@ -0,0 +1,137 @@
...
+}
+
+interface NavLink {
+  to: string;
+  label: string;
+  icon: React.ReactNode;
+  adminOnly?: boolean;
+  /** Match this prefix to mark active (defaults to exact `to`). */
+  match?: string;
+}
```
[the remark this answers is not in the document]
wreiske: Fixed — added `import type { ReactNode } from 'react'` and changed `icon: React.ReactNode` to `icon: ReactNode`.

create-a-container/routers/templates.js
```diff
@@ -0,0 +1,88 @@
...
+  const site = await loadDnsmasqSite(parseInt(req.params.siteId, 10));
+  if (!site) return res.status(404).send('Site not found');
+  res.set('Content-Type', 'text/plain');
+  return res.render(`dnsmasq/${file}`, { site });
+});
+
+router.get('/sites/:siteId/nginx', requireLocalhostOrAdmin, async (req, res) => {
+  const siteId = parseInt(req.params.siteId, 10);
+  const site = await Site.findByPk(siteId, {
+    include: [{
```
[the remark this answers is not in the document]
wreiske: Fixed — added a 404 for missing sites in the nginx route, matching the dnsmasq route.

create-a-container/routers/templates.js
```diff
@@ -0,0 +1,89 @@
...
+            { model: TransportService, as: 'transportService' },
+          ],
+        }],
+      }],
+    }, {
+      model: ExternalDomain,
+      as: 'externalDomains',
+    }],
+  });
+  if (!site) return res.status(404).send('Site not found');
```
runleveldev: This line breaks the bootstrapping assumptions. The nginx template is designed to have a "no site" fallback that allows the administrator to access the API to configure the first site, without that bootstrapping would require plaintext HTTP access, breaking security requirements for registering the first user

create-a-container/package.json
```diff
@@ -1,31 +1,38 @@
 {
   "nodemonConfig": {
     "ignore": [
-      "certs"
+      "certs",
+      "client"
     ]
   },
   "scripts": {
     "dev": "nodemon server.js",
```
runleveldev: Running the server in dev should also run the client in dev, preferably just watching the directory and rebuilding the dist files as needed to prevent the extra http hop that proxying to the dev port from the api server would cause.

create-a-container/client/package.json
```diff
@@ -0,0 +1,34 @@
+{
```
runleveldev, line 1: This package needs to be installed and built in the Makefile in the project root so that the Manager docker container image is setup correctly on first boot. Administrators should not be expected to need to build the client before the UI is available

create-a-container/client/src/app/Sidebar.tsx
```diff
@@ -0,0 +1,142 @@
...
+      label={link.label}
+      icon={link.icon}
+      isActive={isActive(link)}
+      onClick={() => navigate(link.to)}
+    />
+  );
+
+  return (
+    <>
+      <SidebarHeader className="h-16 px-4 py-0">
```
runleveldev, line 122: This loses the previous UI's Site selector (modeled after Portainer) which displays the correct "Containers" (for all users) and "Nodes" (for admins) links depeneding on the last active site. This make navigating the UI a challenge once in a site.

create-a-container/client/src/pages/nodes/NodeImportPage.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+  const { register, handleSubmit, watch, setValue, formState } = useForm<FormData>({
+    resolver: zodResolver(schema),
+    defaultValues: { tlsVerify: true },
+  });
+  const tlsVerify = watch('tlsVerify');
+
+  const mutation = useMutation({
+    mutationFn: (values: FormData) =>
+      api.post<{ imported: number }>(`/api/v1/sites/${siteId}/nodes/import-proxmox`, values),
```
runleveldev: This route is not defined, it appears it should be /import not /import-proxmox

create-a-container/openapi.v1.yaml
```diff
@@ -0,0 +1,449 @@
+openapi: 3.0.3
```
runleveldev, line 1: There was already an OpenAPI spec at create-a-container/openapi.yaml, why was that not just updated?

create-a-container/middlewares/api.js
```diff
@@ -0,0 +1,168 @@
...
+ *   success: { data: <payload>, meta?: { ... } }
+ *   error:   { error: { code: string, message: string, fields?: { [name]: string } } }
+ */
+
+const { doubleCsrf } = require('csrf-csrf');
+
+// --- CSRF (double-submit cookie) ----------------------------------------------------------
+// Token lives in a cookie + must echo in X-CSRF-Token (or _csrf in body).
+// Skipped for Bearer-token API key requests (those are already cryptographically auth'd).
+const isProd = process.env.NODE_ENV === 'production';
```
runleveldev: Session Secrets live in the database as seen in create-a-container/server.js#15. This is to enable secret rotation and state sharing across multiple instances.

create-a-container/middlewares/api.js
```diff
@@ -0,0 +1,168 @@
...
+    secure: isProd,
+    httpOnly: true,
+  },
+  size: 32,
+  getCsrfTokenFromRequest: (req) =>
+    req.headers['x-csrf-token'] || (req.body && req.body._csrf),
+});
+
+// CSRF guard: enforce on state-changing methods, exempt Bearer-auth requests.
+function csrfGuard(req, res, next) {
```
runleveldev: This bypasses CSRF as long as the Authorization: Bearer header is set, regardless of if it's valid, easily worked around by an attacker.

create-a-container/client/src/pages/containers/ContainerFormPage.tsx
```diff
@@ -0,0 +1,519 @@
...
+});
+type FormData = z.infer<typeof schema>;
+
+const COMMON_TEMPLATES = [
+  'ubuntu-22.04-standard',
+  'ubuntu-24.04-standard',
+  'debian-12-standard',
+  'docker:nginx:latest',
+  'docker:postgres:16',
+];
```
runleveldev, line 78: These need to be our provided templates ghcr.io/mieweb/opensource-server/base:latest, ghcr.io/mieweb/opensource-server/nodejs:latest, and ghcr.io/mieweb/ozwell-studio:latest, which are designed explicitly to integrate with this environment. These images technically work but they are not going to be very useful in this environment.

create-a-container/client/src/pages/containers/ContainerFormPage.tsx
```diff
@@ -0,0 +1,519 @@
...
+        setValue('entrypoint', ep);
+      }
+      if (Array.isArray(meta.env)) {
+        meta.env.forEach((e) => {
+          const eq = e.indexOf('=');
+          if (eq > 0) envVars.append({ key: e.slice(0, eq), value: e.slice(eq + 1) });
+        });
+      }
+      setMetadataMsg(`Loaded metadata: ${added} port(s) discovered.`);
+    },
```
runleveldev, line 213: This is not properly loading either the `EXPOSE 22/tcp` port or the custom `LABEL org.mieweb.opensource-server.services.http.default-port=3000` http service from ghcr.io/mieweb/opensource-server/nodejs:latest. Also failed the parse the HTTP service from ghcr.io/mieweb/ozwell-studio:latest.

create-a-container/client/src/pages/containers/ContainerFormPage.tsx
```diff
@@ -0,0 +1,519 @@
...
+                    variant="outline"
+                    leftIcon={<Search className="size-4" />}
+                    isLoading={metadataMutation.isPending}
+                    onClick={() => {
+                      const img = watch('customTemplate') || template;
+                      if (img && img !== 'custom') metadataMutation.mutate(img);
+                    }}
+                  >
+                    Fetch
+                  </Button>
```
runleveldev: The duplication of the input fields for Custom Templates and looking up image metadata is hard to visually parse. "Custom template" seems disabled and mapped to the look up image metadata field, so those should probably just be merged into a single form item.

create-a-container/client/src/pages/auth/RegisterPage.tsx
```diff
@@ -0,0 +1,200 @@
...
+        if (cancelled) return;
+        setInviteError(err instanceof ApiError ? err.message : 'Invitation invalid');
+      } finally {
+        if (!cancelled) setInviteLoading(false);
+      }
+    })();
+    return () => {
+      cancelled = true;
+    };
+  }, [token, reset]);
```
runleveldev, line 73: Unclear why, but I do not see this effect firing in the browser networking tab even though I had a valid token present.

create-a-container/client/src/app/Sidebar.tsx
```diff
@@ -0,0 +1,142 @@
...
+];
+
+export function AppSidebar() {
+  const navigate = useNavigate();
+  const location = useLocation();
+  const { data: session } = useSession();
+  const { isCollapsed, isMobileViewport } = useSidebar();
+  const isAdmin = !!session?.isAdmin;
+  const mfaAdminUrl =
+    isAdmin && session?.pushNotificationUrl ? `${session.pushNotificationUrl}/admin` : null;
```
runleveldev, line 72: I don't see session.pushNotificationUrl being set anywhere, even though I have the URL set in Settings I don't get the Sidebar link.

runleveldev, line 72: Ignore that previous comment on this item. The Sidebar setting is there, but its underneath "Sites" rather than its previous location underneath "Settings"

create-a-container/client/src/pages/apikeys/ApiKeysListPage.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+      {created && (
+        <Alert variant="success">
+          <AlertDescription>
+            <div className="flex flex-col gap-2">
+              <strong>{created.warning}</strong>
+              <div className="flex items-center gap-2">
+                <code className="rounded bg-(--color-surface-2,#f5f5f5) px-2 py-1 font-mono text-sm break-all">
+                  {created.key}
+                </code>
```
runleveldev: This text box is unreadable in darkmode since the background is still white but the text is also white

create-a-container/client/src/pages/containers/ContainersListPage.tsx
```diff
@@ -0,0 +1,188 @@
...
+                            :{h.port}
+                          </span>
+                        ),
+                      )}
+                    </div>
+                  )}
+                </TableCell>
+                <TableCell className="font-mono text-xs">
+                  {c.sshHost && c.sshPort ? `${c.sshHost}:${c.sshPort}` : '—'}
+                </TableCell>
```
runleveldev, line 216: This loses our Open in VSCode (vscode-remote://) and Open in Terminal (ssh://) links

create-a-container/client/src/pages/auth/LoginPage.tsx
```diff
@@ -0,0 +1,389 @@
...
+    formState: { errors, isSubmitting },
+  } = useForm<FormData>({ resolver: zodResolver(schema) });
+
+  useEffect(() => {
+    return () => {
+      if (pollTimer.current) window.clearTimeout(pollTimer.current);
+    };
+  }, []);
+
+  function startPolling(id: string) {
```
runleveldev, line 110: Again not sure why but about 4/5 times the redirect to the next screen fails and sends me back to the login screen instead of on to the login screen. Editing the URL (not just refresshing) will get me the the next screen so it seems auth is working just not the redirect.

create-a-container/routers/api/v1/auth.js
```diff
@@ -0,0 +1,408 @@
...
+
+// POST /api/v1/auth/dev — One-click dev login (non-production only).
+// Creates `dev-admin` or `dev-user` on first use and starts an authenticated
+// session. Disabled when NODE_ENV === 'production'.
+router.post(
+  '/dev',
+  asyncHandler(async (req, res) => {
+    if (process.env.NODE_ENV === 'production') {
+      throw new ApiError(404, 'not_found', 'Not found');
+    }
```
runleveldev: I prefer how this was handled previously, where the route didn't even exist if we were in NODE_ENV production. This "single-button" sign on makes be nervous that itll leak into prod.

runleveldev (changes requested):
Theres more but Im hitting limits of GH API. This PR is so big I had to review from the CLI

create-a-container/server.js
```diff
@@ -123,77 +94,34 @@ async function main() {
...
-    secure: false, // use STARTTLS if supported
-    tls: {
-      rejectUnauthorized: false, // allow self-signed certs
-    },
-  });
-
   // --- Mount Routers ---
-  const loginRouter = require('./routers/login');
-  const registerRouter = require('./routers/register');
-  const verifyRouter = require('./routers/verify');
```
runleveldev, line 150: We still need the /verify route. It's used in the nginx template to enable HTTP Proxy Auth. The current build of this is returning 200 for all requests to /verify effectively disabling it for all containers that are relying on it.

create-a-container/client/src/pages/containers/ContainersListPage.tsx
```diff
@@ -0,0 +1,188 @@
...
+            </TableRow>
+          </TableHeader>
+          <TableBody>
+            {data.map((c: Container) => (
+              <TableRow key={c.id}>
+                <TableCell className="font-medium">{c.hostname}</TableCell>
+                <TableCell>
+                  <Badge variant={statusVariant(c.status)}>{c.status}</Badge>
+                </TableCell>
+                <TableCell>{c.nodeName || '—'}</TableCell>
```
runleveldev: This is supposed to have a link to the {node.apiUrl}/<proxmox terminal ui path> in a new tab.

create-a-container/client/src/pages/containers/ContainerFormPage.tsx
```diff
@@ -0,0 +1,519 @@
...
+                    </div>
+                  )}
+                  {svc.type === 'srv' && (
+                    <Input
+                      label="DNS name"
+                      placeholder="_service._tcp.example"
+                      {...register(`services.${idx}.dnsName`)}
+                    />
+                  )}
+                </div>
```
runleveldev, line 562: The API only supports changing the "Require Auth" flag for existing services. The old UI had an affordance for this by disabling the fields (except for "Delete") and changes had to be made by deleting and adding a new service. If we can support modifying services with this model that's great but the API needs updated to support that or the UI needs updated to indicate it's not supported (by disabling the fields for existing services).

runleveldev (changes requested):

create-a-container/client/src/pages/sites/SiteFormPage.tsx
```diff
@@ -0,0 +1,148 @@
...
+          autoCorrect="off"
+          spellCheck={false}
+          error={formState.errors.internalDomain?.message}
+          hasError={!!formState.errors.internalDomain}
+          {...register('internalDomain')}
+        />
+        <div className="grid gap-4 sm:grid-cols-2">
+          <Input
+            label="DHCP range"
+            placeholder="10.0.0.100-10.0.0.200"
```
runleveldev: This is rendered as is into dnsmasq so it should use a comma as the seperator rather than the -.

create-a-container/client/src/pages/sites/SiteFormPage.tsx
```diff
@@ -0,0 +1,148 @@
...
+        <div className="grid gap-4 sm:grid-cols-2">
+          <Input
+            label="Gateway"
+            placeholder="10.0.0.1"
+            inputMode="numeric"
+            {...register('gateway')}
+          />
+          <Input
+            label="DNS forwarders"
+            placeholder="8.8.8.8 1.1.1.1"
```
runleveldev: Likewise, this is rendered directly into dnsmasq so it should use a comma rather than a space for the seperator.

runleveldev (changes requested):

create-a-container/routers/api/v1/containers.js
```diff
@@ -0,0 +1,580 @@
...
+        if (newHttp.length > 0) {
+          dnsWarnings.push(...(await manageDnsRecords(newHttp, site, 'create')));
+        }
+      }
+    });
+
+    return ok(res, {
+      containerId: container.id,
+      jobId: restartJob ? restartJob.id : null,
+      dnsWarnings,
```
runleveldev, line 524: dnsWarnings are only shown on the ContainerFormPage, but they won't appear because that page redirects back to ContainersListPage where they actually need displayed.

create-a-container/routers/templates.js
runleveldev: 4191d694

create-a-container/package.json
runleveldev: 68cd303a

create-a-container/client/package.json
runleveldev: 10d23159

create-a-container/client/src/app/Sidebar.tsx
runleveldev: ca6c4383

create-a-container/client/src/pages/nodes/NodeImportPage.tsx
runleveldev: 6467b28e

create-a-container/openapi.v1.yaml
runleveldev: 35295102

create-a-container/middlewares/api.js
runleveldev: fb168855

create-a-container/middlewares/api.js
runleveldev: 372a04c6

create-a-container/client/src/pages/containers/ContainerFormPage.tsx
runleveldev: 37628551

create-a-container/client/src/pages/containers/ContainerFormPage.tsx
runleveldev: ba82207a

create-a-container/client/src/pages/containers/ContainerFormPage.tsx
runleveldev: 0364fbe9c4a3

wreiske (commented):
Addressed all unresolved review threads in 1bab2f1. Replies and resolutions to follow per-thread.

create-a-container/client/src/app/Sidebar.tsx
wreiske: Restored in 1bab2f1: when the URL is under `/sites/:siteId/...` the sidebar now shows site-scoped **Containers** (all users) and **Nodes** (admin only) links right under the Sites group, matching the Portainer-style affordance from the prior UI.

create-a-container/routers/templates.js
wreiske: Fixed in 1bab2f1: when no Site exists yet, `/sites/:siteId/nginx` now renders an empty `nginx-conf` (empty `httpServices`/`streamServices`/`externalDomains`) instead of returning 404, so the manager API stays reachable over TLS and the first site can be created without falling back to plaintext HTTP.

create-a-container/client/src/pages/nodes/NodeImportPage.tsx
wreiske: Fixed in 1bab2f1: client now `POST`s to `/api/v1/sites/${siteId}/nodes/import` to match the server route.

create-a-container/client/src/pages/containers/ContainerFormPage.tsx
wreiske: Fixed in 1bab2f1: merged into a single form item. When **Template** = `Custom…` we now show one **Custom image** input with an inline **Fetch metadata** button (and helper text on the format). The duplicate "Or look up image metadata" field is gone.

create-a-container/middlewares/api.js
wreiske: Fixed in 1bab2f1: the bypass now requires **both** an `Authorization: Bearer` header **and** the absence of a session cookie. Browser sessions therefore always go through CSRF; only true API-key clients (which don't carry session cookies) can skip it. An attacker can no longer simply set a bogus `Authorization` header from a session cookie context to bypass.

create-a-container/routers/api/v1/containers.js
wreiske: Fixed in 1bab2f1: on create/edit success the form now passes `dnsWarnings` via `navigate(..., { state: { dnsWarnings } })` to `/sites/:siteId/containers`, and `ContainersListPage` reads `useLocation().state?.dnsWarnings` and renders an `<Alert variant="warning">` at the top of the page. Warnings are now visible on the page the user actually lands on.

create-a-container/client/src/pages/containers/ContainersListPage.tsx
wreiske: Fixed in 1bab2f1: `serializeContainer` now returns `nodeApiUrl`, and `ContainersListPage` wraps the node name in `<a target="_blank" href={`${nodeApiUrl}/#v1:0:=lxc/${vmid}::::`}>` so it deep-links to the Proxmox UI tab for that container.

create-a-container/client/src/pages/auth/RegisterPage.tsx
wreiske: I couldn't reproduce this locally — the effect fires on mount whenever `challengeId` is present. Could you share a HAR / console snippet (or the exact URL you landed on) so I can repro? Happy to dig further once we can pin down the conditions.

create-a-container/client/src/pages/containers/ContainerFormPage.tsx
wreiske: Fixed in 1bab2f1. The previous client code expected a flat `exposedPorts: string[]` shape, which never matched what `GET /api/v1/sites/:siteId/containers/metadata` actually returns from `extractImageMetadata`. The client now consumes the real shape:

- `httpServices[]` (parsed from `LABEL org.mieweb.opensource-server.services.http.*`) are appended as `type=http` services, with `requireAuth` carried over.
- `ports[]` (parsed from `ExposedPorts`, with HTTP service ports filtered out by the server) are appended as `type=tcp`/`type=udp` services — so `EXPOSE 22/tcp` lands as a tcp service on port 22.
- `env` (object), `entrypoint` (string) are populated correctly.

`ContainerMetadata` in `client/src/lib/types.ts` was updated to match.

create-a-container/openapi.v1.yaml
wreiske: Consolidated in 1bab2f1: deleted the legacy `create-a-container/openapi.yaml` (it wasn't referenced anywhere) so `create-a-container/openapi.v1.yaml` is the single source of truth, and the URL versioning (`/api/v1/...`) is explicit in the spec.

create-a-container/client/src/pages/apikeys/ApiKeysListPage.tsx
wreiske: Fixed in 1bab2f1: the displayed API-key `<code>` now uses `bg-zinc-100 text-zinc-900 dark:bg-zinc-800 dark:text-zinc-100`, so it's readable in both themes.

create-a-container/client/src/app/Sidebar.tsx
wreiske: Could you point at which item specifically you mean? The current sidebar has `Sites` as its own (top) group, then an Admin group with `Users / Groups / External Domains / API Keys / Settings`. I want to make sure I move the right one — if it's e.g. `External Domains` (which used to live under Settings) say the word and I'll relocate it.

create-a-container/middlewares/api.js
wreiske: Fixed in 1bab2f1: the CSRF secret is now loaded from the `SessionSecret` table on boot (`initCsrfSecret` in `middlewares/api.js`, awaited from `server.js`) and cached in-process. No env var dependency — rotation and multi-instance sharing now work via the DB row, same as the session signing secret.

create-a-container/client/src/pages/containers/ContainerFormPage.tsx
wreiske: Fixed in 1bab2f1: existing services now render the type `Select` disabled and the hostname / port / DNS / external-domain inputs read-only. Helper text reads *"Existing service — only Require authentication can be changed."* The only editable affordances on an existing service are **Require auth** and **Delete**, matching the old UI's affordance and the API contract.

create-a-container/package.json
wreiske: Updated in 1bab2f1: `npm run dev` now uses `concurrently` to run `nodemon server.js` alongside `vite build --watch`, so `dist/` is rebuilt in place — no extra HTTP hop through the Vite dev server.

create-a-container/client/src/pages/auth/LoginPage.tsx
wreiske: Should be fixed in 1bab2f1: `useLoginMutation`'s `onSuccess` (and the 2FA approved path, and `useDevLoginMutation`) now awaits `qc.refetchQueries({ queryKey: sessionKey })` before resolving. The login awaits the mutation before navigating, so by the time `RequireAuth` mounts on the next route the session query is populated from the server and the guard sees an authenticated user. The "URL works after edit but redirect fails" symptom matches a `useSession` race, and this closes it.

create-a-container/client/src/pages/containers/ContainersListPage.tsx
wreiske: Restored in 1bab2f1: each row in the containers list now has icon buttons that open `vscode://vscode-remote/ssh-remote+${user}@${sshHost}:${sshPort}/` (Open in VSCode) and `ssh://${user}@${sshHost}:${sshPort}` (Open in Terminal).

create-a-container/client/src/pages/sites/SiteFormPage.tsx
wreiske: Fixed in 1bab2f1: placeholder and helper text now read `10.0.0.100,10.0.0.200` with a note that the value is rendered into `dnsmasq` as-is.

create-a-container/routers/api/v1/auth.js
wreiske: Fixed in 1bab2f1: `router.post('/dev', ...)` is now wrapped in `if (process.env.NODE_ENV !== 'production') { ... }`, so the route is not registered at all in production. The client also gates the button behind `useServerInfo().isDev`, but the server is now the authoritative gate.

create-a-container/server.js
wreiske: Restored in 1bab2f1: added `create-a-container/routers/verify.js` mounted at `/verify` (and excluded from the SPA fallback). It accepts session cookies and `Authorization: Bearer <api-key>`, validates the key against `ApiKey` records (using `extractKeyPrefix` + `validateKey`), sets `X-User-ID`, `X-Username`, `X-User-First-Name`, `X-User-Last-Name`, `X-Email`, `X-Groups` identity headers, and returns **200** only on success and **401** otherwise. Containers relying on `auth_request /verify` will behave the same as on `main`.

create-a-container/client/src/pages/sites/SiteFormPage.tsx
wreiske: Fixed in 1bab2f1: forwarders helper text now reads `8.8.8.8,1.1.1.1` (comma) and notes that it is rendered into `dnsmasq` as-is.

create-a-container/client/package.json
wreiske: Added in 1bab2f1: the root `Makefile`'s `install-create-container` target now runs `cd create-a-container/client && npm install && npm run build`, so the Manager image has `dist/` available on first boot without administrator intervention.

create-a-container/client/src/pages/containers/ContainerFormPage.tsx
wreiske: Fixed in 1bab2f1: `COMMON_TEMPLATES` now lists only the mieweb-provided images — `ghcr.io/mieweb/opensource-server/base:latest`, `ghcr.io/mieweb/opensource-server/nodejs:latest`, and `ghcr.io/mieweb/ozwell-studio:latest`. The legacy `ubuntu-*` / `debian-12-standard` / generic-docker entries were removed.

create-a-container/client/src/pages/containers/ContainerFormPage.tsx
```diff
@@ -0,0 +1,530 @@
...
+            />
+            {!isEdit && (
+              <>
+                <Select
+                  label="Template"
+                  value={template || ''}
+                  onValueChange={(v) => setValue('template', v)}
+                  options={templateOptions}
+                />
+                {template === 'custom' && (
```
runleveldev: This looks bad, the button is misaligned.

Image

create-a-container/client/src/app/Sidebar.tsx
runleveldev: I was refering to the "MFA Admin" external link sourced from the Push Notification API URL system setting. It's hidden unless that setting is set and it should be rendered as part of the Admin group.

create-a-container/client/src/app/Sidebar.tsx
runleveldev: This is still missing the current UI dropdown select which allows a user to change sites in the sidebar without needing to vist the sites page to select a new site. It also maintains the "current site" state as the user navigates to non-site pages (such as setting or users) so they can quickly navigate back to where they were.

create-a-container/client/src/pages/containers/ContainerFormPage.tsx
```diff
@@ -0,0 +1,530 @@
...
+              error={formState.errors.hostname?.message}
+              hasError={!!formState.errors.hostname}
+              {...register('hostname')}
+            />
+            {!isEdit && (
+              <>
+                <Select
+                  label="Template"
+                  value={template || ''}
+                  onValueChange={(v) => setValue('template', v)}
```
runleveldev: When a template is selected, even one of the default ones, the metadata needs to be loaded into the form the same way as if (in this pr) the "Fetch Metadata" button is selected for a custom template. We're expecting the frontend to fill these values automatically on the current UI so this would break user workflows.

create-a-container/client/src/pages/containers/ContainerFormPage.tsx
```diff
@@ -0,0 +1,530 @@
...
+  const [metadataMsg, setMetadataMsg] = useState<string | null>(null);
+  const metadataMutation = useMutation({
+    mutationFn: (image: string) => queries.containerMetadata(siteId!, image),
+    onSuccess: (meta: ContainerMetadata) => {
+      let added = 0;
+      (meta.httpServices || []).forEach((svc) => {
+        services.append({
+          type: 'http' as FormData['services'][number]['type'],
+          internalPort: String(svc.port),
+          externalHostname: '',
```
runleveldev: The current UI defaults externalHostname to the name of the container (as specified in the 'hostname' form field). Unless the httpService has the 'hostnameSuffix' field in which case it should default to `{hostname}-{hostnameSuffix}`.

create-a-container/client/src/pages/containers/ContainerFormPage.tsx
```diff
@@ -0,0 +1,530 @@
...
+          externalDomainId: '',
```
runleveldev: The logic the current frontend uses is: use the first external domain that has the current site as default. If the current site as no domains with it as default (to be clear, a domain has a default site, so a single site may have many default domains) then just pick the first domain. We're trying to avoid expecting users to have to decide the domain for themselves.

create-a-container/client/src/pages/containers/ContainerFormPage.tsx
```diff
@@ -0,0 +1,530 @@
...
+                  <p className="text-xs text-muted-foreground">{metadataMsg}</p>
+                )}
+              </>
+            )}
+            <Input
+              label="Entrypoint"
+              placeholder="Override container entrypoint"
+              {...register('entrypoint')}
+            />
+            {bootstrap?.nvidiaAvailable && (
```
runleveldev: Per the current UI, I want the NVIDIA switch to always be visible, just not enabled for sites that don't have NVIDIA support. The current UI has a tooltip explaining no NVIDIA nodes are available when hovering over the disabled checkbox.

create-a-container/client/src/pages/containers/ContainerFormPage.tsx
```diff
@@ -0,0 +1,530 @@
...
+                    <div className="grid gap-3 sm:grid-cols-2">
+                      <Input
+                        label="External hostname"
+                        placeholder="app"
+                        readOnly={isExisting}
+                        {...register(`services.${idx}.externalHostname`)}
+                      />
+                      <Select
+                        label="External domain"
+                        value={svc.externalDomainId || ''}
```
runleveldev, line 538: Per previous comment, the current UI has selection logic for picking a domain for the user rather than defaulting to requiring the user to pick one.

create-a-container/client/src/pages/containers/ContainerFormPage.tsx
runleveldev: The current UI has "user friendly" names: "Debian 13", "NodeJS 24" and "Ozwell Studio".

create-a-container/client/src/pages/auth/RegisterPage.tsx
runleveldev: I generated this URL in my local compose setup https://manager.localhost/register?token=f02cec91f7f62b3b1c0b82aa5fde67dba51b5679f752cae4202e66edb965fe41 obviously you can't hit my localhost so here's the HAR

[manager.localhost_register_Archive \[26-05-26 16-34-28\].har.json](https://github.com/user-attachments/files/28283548/manager.localhost_register_Archive.26-05-26.16-34-28.har.json)

create-a-container/client/src/lib/api.ts
```diff
@@ -0,0 +1,146 @@
...
+    });
+  };
+
+  let response = await doFetch();
+
+  // CSRF token may have rotated; retry once.
+  if (response.status === 403 && isMutation) {
+    const text = await response.clone().text();
+    if (text.includes('csrf') || text.includes('CSRF')) {
+      await fetchCsrfToken();
```
runleveldev: Only fetching a csrf token when the response status is 403 is generating a lot of network requests. This clutters logs (both server and broweser) but more importantly causes the express-rate-limit setting to be reached (which shouldn't happen at all under normal operation). Express is currently only rate limiting errors but doing so fairly strictly.

According to csrf-csrf, since we're using express-session, we should be using csrf-sync instead which stores the csrf token as part of the session state. This would also remove the need for the session secret synchronization from my previous review comment.

>Sessions
>
>Once again, if you are using sessions, you should be using [csrf-sync](https://github.com/Psifi-Solutions/csrf-sync) instead. Sessions have server side state by default.

create-a-container/client/src/pages/auth/LoginPage.tsx
runleveldev: I'm seeing this now. It's still ~ 4 out of every 5 times then I sometimes get in as expected.

Image

It seems its actually logging me in successfully, because I can then navigate directly to /sites and get the logged-in UI, but refreshing the page doesn't trigger the redirect like it would have on the old UI if you had a valid session cookie but had mistakenly gone to the login page.

Essentially I'm asking for 2 fixes. (1) fix that error that's been scrennshotted. (2) if the user navigates to /login when they already have a valid session, immediately redirect them to the redirect query parameter or / if they didn't provide ?redirect

runleveldev:
@wreiske Doug didn't think you'd have the cycles to finish this PR so I was fixing it up in a local branch (that was the random letters I was commenting, references to local commits I had). If you want to take the ball on seeing this through I'll let you but let me know soon which way you'd rather lean so I know I need to start implementing the fixes instead.

create-a-container/client/src/pages/auth/RegisterPage.tsx
runleveldev: This is the contents of my "InviteTokens" table as well you may be able to seed the data rather than trying to setup an email server (I've been sending out smtp-lb over VPN but that's probably not a great idea for development usecase).

```
                  id                  |        email        |                              token                               |         expiresAt          | used |         createdAt          |         updatedAt
--------------------------------------+---------------------+------------------------------------------------------------------+----------------------------+------+----------------------------+----------------------------
 07d69809-f681-4390-a9ef-ff0668565d6d | rgingras@mieweb.com | f02cec91f7f62b3b1c0b82aa5fde67dba51b5679f752cae4202e66edb965fe41 | 2026-05-27 20:31:14.272+00 | f    | 2026-05-26 20:31:14.273+00 | 2026-05-26 20:31:14.273+00
```

create-a-container/package.json
```diff
@@ -1,35 +1,45 @@
...
     "pg": "^8.16.3",
     "qrcode": "^1.5.4",
     "sequelize": "^6.37.8",
     "sequelize-cli": "^6.6.3",
+    "sqlite3": "^6.0.1",
     "swagger-ui-express": "^5.0.1",
     "yamljs": "^0.3.0"
   },
   "devDependencies": {
+    "concurrently": "^9.1.0",
```
runleveldev, line 42: package-lock.json wasn't updated with this so `npm ci` is currently failing.

runleveldev (changes requested):

create-a-container/package.json
runleveldev: 8e9f0608

create-a-container/client/src/lib/api.ts
runleveldev: 37d339d3

create-a-container/client/src/pages/containers/ContainerFormPage.tsx
runleveldev: 36cef7681d3c

create-a-container/client/src/pages/containers/ContainerFormPage.tsx
runleveldev: 36cef7681d3c

create-a-container/client/src/pages/containers/ContainerFormPage.tsx
runleveldev: af0215a0c90d

create-a-container/client/src/pages/containers/ContainerFormPage.tsx
runleveldev: 6a9b0c2003d1

create-a-container/client/src/pages/containers/ContainerFormPage.tsx
runleveldev: 22ea873be9fe

create-a-container/client/src/pages/containers/ContainerFormPage.tsx
runleveldev: 22ea873be9fe

create-a-container/client/src/app/Sidebar.tsx
runleveldev: 16ea792babcb

create-a-container/client/src/pages/containers/ContainerFormPage.tsx
runleveldev: 6ea45d48e6d2

create-a-container/client/src/app/Sidebar.tsx
runleveldev: 3f85bb4425d7

create-a-container/client/src/pages/auth/RegisterPage.tsx
runleveldev: e812e38d7527

runleveldev (approved):

cmyers-mieweb:
I found some obscure bugs that occur when rapidly moving between pages that will crash the webpage, but we can merge now and address those in a separate PR since it is difficult to replicate. I'm also going to work on some additional UI/UX improvements. 
image

PR #329, opened 2026-06-03
Run the Manager locally without Proxmox (make dev + mock provisioning)

horner:
## Why

Until now, the only way to exercise the **Manager** (`create-a-container`) web app was the full Docker stack (Proxmox, provisioning, DNS, reverse-proxy). That is slow to spin up and overkill for iterating on the web app itself. There was no way to:

- Run the Manager standalone against SQLite, and
- Actually **create a container** without a real Proxmox node.

This PR makes the Manager fully usable on `localhost` for web-app development, with a clear hook point for a future Docker-based local provisioner.

### 1. `make dev` — run the Manager standalone (SQLite)
- New `dev` target sets up and starts `create-a-container` against local SQLite: creates `.env` (with a generated `SESSION_SECRET`) if missing, installs deps, runs migrations, builds the client, and serves at `http://localhost:3000`.
- Fixed `db:migrate` script to call `sequelize-cli` (the `sequelize` bin alias isn't installed).

### 2. On-demand local environment on dev login
The `/dev` login route now self-provisions everything the new-container form needs, using `findOrCreate` (idempotent, dev-only):
- a `localhost` **Site**,
- a **local Node** (`apiUrl: 'local'`) that acts as the mock hypervisor,
- a default **External Domain** (`localhost`).

The external domain is required because image templates that expose an HTTP port auto-add an HTTP service, and the form's validation requires a domain for HTTP services. Without it the new-container form was silently unsubmittable.

### 3. Mock provisioning for the local node
In `POST /containers`, when running in dev against the `local` node (`apiUrl === 'local'`), Proxmox provisioning is short-circuited: the container is immediately marked `running` with a placeholder VMID/MAC/IP. This is the documented hook point where a real Docker-based local provisioner can plug in later.

### 4. Bug fix — erroneous `UNIQUE` on `Containers.nodeId`
SQLite's `changeColumn` rebuilds the table and accidentally applied a **column-level `UNIQUE`** on `Containers.nodeId`, which limited each node to a single container. A node (a Proxmox host) owns many containers; the correct uniqueness guarantee is already the `(nodeId, containerId)` composite index. New migration drops the bad constraint. **No-op on Postgres**, which was never affected (its `ALTER COLUMN` doesn't rebuild tables).

### 5. Docs
- New "Run the Manager Locally (`make dev`)" section in the development workflow doc, with the Docker full-stack content demoted to a subsection.
- README + copilot-instructions point at that doc as the single source of truth.

## Testing
Verified end-to-end in the browser: dev login auto-provisions site/node/domain, then created multiple containers (Debian 13, NodeJS 24 with an HTTP service) — all land as `running` with unique VMID/IP and no Proxmox required.

### DIFF
--- a/create-a-container/.gitignore
+++ b/create-a-container/.gitignore
@@ -1,3 +1,4 @@
 .env
 node_modules
 data/
+*.sqlite
--- a/create-a-container/migrations/20260603000000-remove-node-id-unique-from-containers.js
+++ b/create-a-container/migrations/20260603000000-remove-node-id-unique-from-containers.js
@@ -1,0 +1,86 @@
+'use strict';
+
+/**
+ * SQLite's changeColumn recreates the table and accidentally applied a column-
+ * level UNIQUE on Containers.nodeId.  A node (Proxmox host) can own many
+ * containers, so the constraint is wrong.  The correct uniqueness guarantee is
+ * already enforced by the (nodeId, containerId) composite index.
+ *
+ * Postgres was never affected because its ALTER COLUMN does not rebuild tables.
+ * This migration is a no-op on Postgres (removeConstraint on a non-existent
+ * constraint will throw, so we guard).
+ */
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    const dialect = queryInterface.sequelize.getDialect();
+
+    if (dialect === 'sqlite') {
+      // SQLite requires a full table rebuild to drop a column-level constraint.
+      await queryInterface.sequelize.transaction(async (t) => {
+        // 1. Create replacement table without the UNIQUE on nodeId
+        await queryInterface.sequelize.query(
+          `CREATE TABLE "Containers_new" AS SELECT * FROM "Containers" WHERE 0`,
+          { transaction: t },
+        );
+        await queryInterface.sequelize.query(`DROP TABLE "Containers_new"`, { transaction: t });
+
+        // Recreate properly using Sequelize's createTable so the model sync
+        // matches.  Easier: use raw DDL mirroring the current schema minus UNIQUE.
+        const [[{ sql }]] = await queryInterface.sequelize.query(
+          `SELECT sql FROM sqlite_master WHERE type='table' AND name='Containers'`,
+          { transaction: t },
+        );
+
+        // Replace the column-level UNIQUE on nodeId
+        const fixed = sql.replace(
+          /`nodeId` INTEGER NOT NULL UNIQUE/,
+          '`nodeId` INTEGER NOT NULL',
+        );
+
+        if (fixed === sql) {
+          // Constraint not present — nothing to do
+          return;
+        }
+
+        const newName = 'Containers_new';
+        await queryInterface.sequelize.query(
+          fixed.replace(/CREATE TABLE "Containers"/, `CREATE TABLE "${newName}"`),
+          { transaction: t },
+        );
+        await queryInterface.sequelize.query(
+          `INSERT INTO "${newName}" SELECT * FROM "Containers"`,
+          { transaction: t },
+        );
+        await queryInterface.sequelize.query(`DROP TABLE "Containers"`, { transaction: t });
+        await queryInterface.sequelize.query(
+          `ALTER TABLE "${newName}" RENAME TO "Containers"`,
+          { transaction: t },
+        );
+
+        // Recreate the indexes that were on the old table
+        await queryInterface.sequelize.query(
+          `CREATE UNIQUE INDEX IF NOT EXISTS "containers_node_id_container_id_unique" ON "Containers" ("nodeId", "containerId")`,
+          { transaction: t },
+        );
+        await queryInterface.sequelize.query(
+          `CREATE UNIQUE INDEX IF NOT EXISTS "containers_site_hostname_unique_idx" ON "Containers" ("siteId", "hostname")`,
+          { transaction: t },
+        );
+        await queryInterface.sequelize.query(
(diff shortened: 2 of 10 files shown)


create-a-container/migrations/20260603000000-remove-node-id-unique-from-containers.js
runleveldev, line 1: We are constantly fighting this bug with sqlite which is why production and the current docker compose stack use postgresql. I'm trying to avoid merging sqlite specific fixes with plans to remove sqlite support in a future PR. Officially Postgres is the only supported DB.

create-a-container/routers/api/v1/auth.js
```diff
@@ -68,6 +70,29 @@ if (process.env.NODE_ENV !== 'production') {
       const isAdmin = req.body?.role === 'admin';
       const uid = isAdmin ? 'dev-admin' : 'dev-user';
 
+      // Ensure a localhost site exists to work with.
```
runleveldev, line 73: These new defaults look like they're going to conflict with the bootstrapping done by the Docker Compose setup. I'll need to be at my laptop to test it though.

create-a-container/routers/api/v1/containers.js
```diff
@@ -347,6 +348,30 @@ router.post(
         }
       }
 
+      // Local dev: short-circuit Proxmox provisioning. The on-demand `local`
+      // node (apiUrl === 'local') has no real hypervisor, so mark the container
+      // running immediately with placeholder VMID/MAC/IP. This is the hook point
+      // where a real Docker-based local provisioner can plug in later.
+      if (process.env.NODE_ENV !== 'production' && node.apiUrl === 'local') {
```
runleveldev, line 355: I don't think I can express enough how much of a bad idea I think this is. If changes are hitting this code path, they need to be QAed against a real Proxmox API. That's the whole point of the docker-compose setup. I have always taken a hard stance against mocks because its impossible to guarantee this is going to even remotely replicate what the actual code path is doing. Adding this will make my job as a code review significantly harder going forward because I will have no assurance that changes were ever tried against a Proxmox environment. I'm willing to entertain a lighterweight WebUI development workflow, but I won't approve mock providers since it drastically changes the assumptions about the environment.

create-a-container/.gitignore
```diff
@@ -1,3 +1,4 @@
 .env
 node_modules
 data/
+*.sqlite
```
runleveldev, line 4: Why is this here if its already in the root .gitignore?

create-a-container/routers/api/v1/auth.js
```diff
@@ -68,6 +70,29 @@ if (process.env.NODE_ENV !== 'production') {
...
+      const [site] = await Site.findOrCreate({
+        where: { name: 'localhost' },
+        defaults: { internalDomain: 'localhost', externalIp: '127.0.0.1' },
+      });
+
+      // Ensure a local dev node exists so containers can be created without
+      // a real Proxmox host. `apiUrl: 'local'` marks it as the mock node;
```
runleveldev, line 80: Just to reiterate my previous comment, I don't like mocks, but this implementation relying on in-band signaling is a massive anti-pattern since it overloads the use of that database field creating a convention contributors have to "just know" to understand why their mock is or isn't working. It's not really self-documenting. If I'm going to have to merge mock support, at least give me a dedicated field like 'dummyNode' (boolean) which only gets used if there's no real nodes available.

create-a-container/routers/api/v1/containers.js
runleveldev: As a follow-up, I would be happier with this if there was an abstraction layer instead of a dedicated code path. Currently we just have the ProxmoxApi class, but I envision that providing an interface (NodeApi) which could then be implemented by a DummyApi class that the mock node uses. Code that requires interacting with Nodes would call through the appropriate Api class (determined by a new nodeType field). This would still require running the job-runner.js alongside the server if you need containers to get created, but that's a benefit to me since it ensures the full container creation code path if being exercised.

runleveldev (changes requested):
If it's not obvious per my other comments, I don't like this strategy. I understand the desire to rapidly iterate the web UI but with how heavily integrated this is to its hosting environment I don't believe any meaningful changes can be created without the ability to try them out against a real environment. The docker-compose workflow was intentionally designed to make doing so as painless as possible. In addition to my other comments, I believe supporting multiple development workflows will result in drift between team members and new contributors over time which will not happen only supporting the docker-compose workflow. Likewise, docker-compose ensures that development happens in a standard environment so the application is not accidentally made dependent on local development tools nor does it need code branches for alternate tool implementations or architectures (think awk flags on mac vs linux).

I reference these in my other comments, but as a summary, I will not approve this PR without the following changes:

1. Postgres as the primary DB. Everyone has docker installed this should not be an issue
2. Out-of-band signaling that a Node is mock preferably via a new field 'type' in the Node model set to 'dummy' (or similar) for a mock node
3. An abstraction on the ProxmoxApi class allowing dummy nodes to be interacted with using the same code paths as Proxmox nodes

Even with those I still don't like the idea of mock APIs (how do you know your changes work against real APIs if you don't try them out?) but I will at least be okay with a unified implementation that doesn't require per-environment code paths.

create-a-container/routers/api/v1/auth.js
runleveldev: I confirmed this messed up the docker-compose bootstrap. It imports the nodes and containers into this localhost site instead of the intended "Development" site which has the IP ranges configured properly.

runleveldev:
superceding with #353

PR #332, opened 2026-06-04
Resource Request Feature

### LINKED ISSUE #258: Add ability to choose a size of the virtual env size
## Request

Image

## Naming convention
Here is a group chat: https://chatgpt.com/gg/v/69c7ea46b47481a0bfd888f5fa6e1714?token=ljKKKiaPHiWD_93XgIUI4g about the naming convention.

Shape | vCPU | Memory | Use Case
-- | -- | -- | --
nano | 0.25 | 0.5 GB | sidecars, tiny workers
micro | 0.5 | 1 GB | lightweight APIs
small | 1 | 2 GB | standard services
medium | 1 | 4 GB | memory-heavy APIs
large | 2 | 8 GB | backend services
xlarge | 4 | 16 GB | compute + memory
2xlarge | 8 | 32 GB | heavy processing

## Temporary Request

For plain Linux containers, memory is enforced with cgroups, so the limit on a running container can be changed by the runtime or orchestrator

### Commands in the container

```
mieos request 1h ram 32gb

mieos request 3d disk 30gb

mieos request 30m cpu +3
 
```
These are examples of requested in a temporary burst of 1h of ram, 3 days of disk, and 30 minutes of cpu.

If after the time expires and the resources cannot be released, the owner of the container will be bugged by the hour for 24 hours and then the container will just be restarted to release the resources.

cmyers-mieweb:
Resolves: https://github.com/mieweb/opensource-server/issues/258

This pull request introduces a new resource request system for container resources, allowing users to request changes to resource allocations (such as RAM, CPUs, etc.) for their containers, with an approval workflow for changes above default values. It adds new UI components for managing and requesting resources, updates navigation and routing to support new resource request pages, and expands the API query and type definitions to support these features.

**Resource Request System Implementation:**

* Added a new `ResourcesSection` React component (`ResourcesSection.tsx`) that displays and allows editing of container resources, with a workflow for submitting resource requests and handling admin approval. This component is integrated into the container form page. [[1]](diffhunk://#diff-db1a34fba901c420bf64ec383e0797de443128323a779338c9540f5b7cfef23eR1-R238) [[2]](diffhunk://#diff-94ee4a3b78795dc7d8b17b4066f261aabd1e48c80b251c9bb8e95f9e9989846aR579-R587)
* Defined new types `ResourceRequest`, `EffectiveResources`, and `ResourceType` in `types.ts` for strong typing of resource requests and allocations.

**API and Query Enhancements:**

* Extended the queries and keys in `queries.ts` to support listing resource requests, counting pending requests, and fetching effective resources for a container. [[1]](diffhunk://#diff-3ec7271b4d8a874c93fe3877afeab9bc537844a6e1ecc0703884d0257f2e7b82R45-R48) [[2]](diffhunk://#diff-3ec7271b4d8a874c93fe3877afeab9bc537844a6e1ecc0703884d0257f2e7b82R97-R108)
* Updated imports and type usages throughout the codebase to support these new features.

**Navigation and Routing Updates:**

* Updated the sidebar navigation to include links for "Resource Requests" (for admins, with a pending count badge) and "My Requests" (for regular users). [[1]](diffhunk://#diff-73dec8f574703ef304f60c834bfb19daa296479389311d20601a16f9261dae84R18) [[2]](diffhunk://#diff-73dec8f574703ef304f60c834bfb19daa296479389311d20601a16f9261dae84R72-R79) [[3]](diffhunk://#diff-73dec8f574703ef304f60c834bfb19daa296479389311d20601a16f9261dae84R200-R217)
* Added routes and page imports for the new resource request pages (`ResourceRequestsPage` and `MyRequestsPage`). [[1]](diffhunk://#diff-65e8c60e95f1cfe5f50137f11eb75479913707599687553c516efc586e85120aR27-R28) [[2]](diffhunk://#diff-65e8c60e95f1cfe5f50137f11eb75479913707599687553c516efc586e85120aR83-R85)

**Miscellaneous Improvements:**

* Improved `.dockerignore` to exclude all `node_modules` directories recursively.
* Added a debounced value hook for the hostname field in the container form to optimize resource queries. [[1]](diffhunk://#diff-94ee4a3b78795dc7d8b17b4066f261aabd1e48c80b251c9bb8e95f9e9989846aR27-R38) [[2]](diffhunk://#diff-94ee4a3b78795dc7d8b17b4066f261aabd1e48c80b251c9bb8e95f9e9989846aR137)

These changes lay the groundwork for a more robust and user-friendly resource management workflow in the application.

rr5
rr4
rr3
rr2
rr1

### DIFF
--- a/create-a-container/bin/create-container.js
+++ b/create-a-container/bin/create-container.js
@@ -28,7 +28,7 @@ const path = require('path');
 
 // Load models from parent directory
 const db = require(path.join(__dirname, '..', 'models'));
-const { Container, Node, Site, Service, HTTPService, ExternalDomain, Setting } = db;
+const { Container, Node, Site, Service, HTTPService, ExternalDomain, Setting, ResourceRequest } = db;
 
 // Load utilities
 const { parseArgs } = require(path.join(__dirname, '..', 'utils', 'cli'));
@@ -182,7 +182,19 @@ async function main() {
   console.log(`Node: ${node.name}`);
   console.log(`Site: ${site.name} (${site.internalDomain})`);
   console.log(`Template: ${container.template}`);
-  
+
+  // Look up approved resource requests for this container identity
+  const approvedResources = await ResourceRequest.getApprovedResources(
+    site.id,
+    container.hostname,
+    container.username,
+  );
+  const cores = approvedResources.cpus || 4;
+  const memory = approvedResources.memory || 4096;
+  const swap = approvedResources.swap !== undefined ? approvedResources.swap : 0;
+  const rootfsSize = approvedResources.rootfs || 50;
+  console.log(`Resources: cores=${cores}, memory=${memory}MB, swap=${swap}MB, rootfs=${rootfsSize}GB`);
+
   const isDocker = isDockerImage(container.template);
   console.log(`Template type: ${isDocker ? 'Docker image' : 'Proxmox template'}`);
   
@@ -248,16 +260,16 @@ async function main() {
         hostname: container.hostname,
         ostemplate,
         description: `Created from Docker image ${container.template}`,
-        cores: 4,
+        cores,
         features: 'nesting=1,keyctl=1,fuse=1',
-        memory: 4096,
+        memory,
         net0: `name=eth0,ip=dhcp,bridge=${node.networkBridge}`,
         searchdomain: site.internalDomain,
-        swap: 0,
+        swap,
         onboot: 1,
         tags: container.username,
         unprivileged: 1,
-        rootfs: `${rootfsStorage}:50`
+        rootfs: `${rootfsStorage}:${rootfsSize}`
       });
       console.log(`Create task started: ${createUpid}`);
       
@@ -298,12 +310,12 @@ async function main() {
       // Configure the container (Docker containers are configured at creation time)
       console.log('Configuring container...');
       await client.updateLxcConfig(node.name, vmid, {
-        cores: 4,
+        cores,
         features: 'nesting=1,keyctl=1,fuse=1',
-        memory: 4096,
+        memory,
         net0: `name=eth0,ip=dhcp,bridge=${node.networkBridge}`,
         searchdomain: site.internalDomain,
-        swap: 0,
+        swap,
         onboot: 1,
         tags: container.username
       });
--- a/create-a-container/bin/create-container.js
+++ b/create-a-container/bin/create-container.js
@@ -191,7 +191,7 @@ async function main() {
   );
   const cores = approvedResources.cpus || 4;
   const memory = approvedResources.memory || 4096;
-  const swap = approvedResources.swap !== undefined ? approvedResources.swap : 0;
+  const swap = approvedResources.swap || 0;
   const rootfsSize = approvedResources.rootfs || 50;
   console.log(`Resources: cores=${cores}, memory=${memory}MB, swap=${swap}MB, rootfs=${rootfsSize}GB`);
 
--- a/create-a-container/migrations/20260602120000-create-resource-requests.js
+++ b/create-a-container/migrations/20260602120000-create-resource-requests.js
@@ -1,0 +1,81 @@
+'use strict';
+
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    await queryInterface.createTable('ResourceRequests', {
+      id: {
+        allowNull: false,
+        autoIncrement: true,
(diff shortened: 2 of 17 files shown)


create-a-container/bin/create-container.js
```diff
@@ -182,7 +182,19 @@ async function main() {
...
+
+  // Look up approved resource requests for this container identity
+  const approvedResources = await ResourceRequest.getApprovedResources(
+    site.id,
+    container.hostname,
+    container.username,
+  );
+  const cores = approvedResources.cpus || 4;
+  const memory = approvedResources.memory || 4096;
+  const swap = approvedResources.swap !== undefined ? approvedResources.swap : 0;
```
runleveldev: Why isn't this using the same <resource || default> pattern as the previous and following assignments?

create-a-container/migrations/20260602120000-create-resource-requests.js
```diff
@@ -0,0 +1,81 @@
...
+      },
+      hostname: {
+        type: Sequelize.STRING(255),
+        allowNull: false,
+      },
+      username: {
+        type: Sequelize.STRING(255),
+        allowNull: false,
+      },
+      requestedBy: {
```
runleveldev, line 27: Is username and requestedBy not the same thing?

create-a-container/routers/api/v1/resource-requests.js
```diff
@@ -0,0 +1,265 @@
...
+  asyncHandler(async (req, res) => {
+    const request = await ResourceRequest.findByPk(parseInt(req.params.id, 10));
+    if (!request) throw new ApiError(404, 'not_found', 'Request not found');
+    if (request.status !== 'pending') {
+      throw new ApiError(409, 'already_reviewed', 'Request has already been reviewed');
+    }
+
+    // Verify at least one provisioned container exists for this request before approving.
+    // A container is considered provisioned when it has a Proxmox VMID (containerId is not null).
+    const provisionedContainer = await Container.findOne({
```
runleveldev, line 157: I think this check should be moved after setting the request to approved in the database but before applying it to any running container. A user should be able to pre-load resource requests if they know they're going to make a container that will need them.

create-a-container/routers/api/v1/resource-requests.js
```diff
@@ -0,0 +1,265 @@
...
+    notifyRequestReviewed(request).catch((err) =>
+      console.warn('Failed to send resource request denial email:', err.message),
+    );
+
+    return ok(res, request);
+  }),
+);
+
+/**
+ * Send an approval/denial email to the user who submitted the request.
```
runleveldev: This scares me a little bit. I don't want to spam people with emails since we don't yet have a way to opt-out. Our domain reputation is still building I don't want to risk spam reports for non-critical notifications.

runleveldev (changes requested):

create-a-container/routers/api/v1/resource-requests.js
cmyers-mieweb: The reason why this check was included was of a bug I found where if a user requests extra resources during the build form, but does NOT proceed to submit to the container creator and backs out, a request will be received and if approved, the page will break.

create-a-container/routers/api/v1/resource-requests.js
cmyers-mieweb: I'll remove the email feature

create-a-container/bin/create-container.js
cmyers-mieweb: I've modified to keep the consistency

create-a-container/migrations/20260602120000-create-resource-requests.js
cmyers-mieweb: They have the same value, but kept it seperated in the migration for ease of identification. I am creating a new migration to drop this in favor username.

create-a-container/routers/api/v1/resource-requests.js
runleveldev: That's fine, but the reason I wanted requests to be handled like this (ie. based on name and not tied to a specific container) is so that they can be pre-loaded and also persist across container rebuilds. This check prevents the admin from approving outright if the container doesn't exist. I think if you just move the saving of the approval to the database above this check it should work as expected.

cmyers-mieweb:
added site to request 
image

runleveldev (approved):

PR #340, opened 2026-06-09
Integrate NetBox VM management

cmyers-mieweb:
Add NetBox integration to register and remove LXC records that mirror managed containers.

- Add `utils/netbox.js` implementing NetBox API helpers: `nbFetch`, `findClusterId`, `createVirtualMachine`, `deleteVirtualMachine`, and withNetbox (loads netbox_url/netbox_token from Settings).
- Use withNetbox/createVirtualMachine in `bin/create-container.js` to register a VM after container creation (non-fatal; errors are warned).
- Use withNetbox/deleteVirtualMachine in routers/api/v1/containers.js to remove the VM when a container is deleted.
- Expose new settings keys (netbox_url, netbox_token) in `routers/api/v1/settings.js` for GET/PUT so NetBox credentials can be configured.

The NetBox deletion helper is resilient to errors (logs but does not throw) to avoid blocking core workflows.

## NetBox Integration — How It Works

All NetBox logic lives in utils/netbox.js. It's invoked from two places: bin/create-container.js (on create) and routers/api/v1/containers.js (on delete).

---

### Prerequisites (manual setup in NetBox)

The integration is **opt-in** — it silently does nothing if these two settings are absent from the `Settings` table:

| Setting key | Value |
|---|---|
| `netbox_url` | Base URL, e.g. `https://netbox.example.com` |
| `netbox_token` | API token with write access to IPAM + Virtualization |

The integration also currently expects a **cluster** to already exist in NetBox whose name matches your `Site.name`. The cluster is required; site and device are looked up opportunistically.

---

### On container creation (create-container.js)

Triggered after the container is running and has an IP address. Steps:

```mermaid
flowchart TD
    A[withNetbox: check settings] -->|missing| Z[skip silently]
    A -->|present| B[Parallel lookups]
    B --> C[findClusterId — required\nmatches Site.name]
    B --> D[findSiteId — optional\nmatches Site.name in /dcim/sites/]
    B --> E[findDeviceId — optional\nmatches Node.name in /dcim/devices/]
    C --> F[POST /virtualization/virtual-machines/\nname=hostname, cluster, site?, device?\ncomments = created-by username]
    F --> G[POST /virtualization/interfaces/\neth0 on the VM]
    G --> H[POST /ipam/ip-addresses/\nassigned to eth0, CIDR notation]
    H --> I[PATCH virtual-machine\nset primary_ip4]
```

- If `findClusterId` fails (cluster not found), the whole call throws — caught by the `try/catch` in create-container.js and logged as a **non-fatal warning**. The container is still marked `running`.
- `site` and `device` are only included in the VM body if the lookups return a result. Missing ones are silently omitted.

---

### On container deletion (containers.js DELETE handler)

```mermaid
flowchart TD
    A[withNetbox: check settings] -->|missing| Z[skip silently]
    A -->|present| B[GET /virtualization/virtual-machines/?name=hostname]
    B -->|not found| C[log and return]
    B -->|found| D[GET /virtualization/interfaces/?virtual_machine_id=id]
    D --> E[For each interface:\nGET assigned IP addresses]
    E --> F[DELETE each IP address]
    F --> G[DELETE each interface]
    G --> H[DELETE virtual-machine]
```

Deletion is **fully non-throwing** — any error is caught, logged to stderr, and the primary container deletion proceeds regardless.

---

### What ends up in NetBox per container

| NetBox object | Value |
|---|---|
| `virtualization.virtual-machine` | `name` = hostname, `cluster` = site name, `site` = site (if found), `device` = Proxmox node (if found), `comments` = `"This container was built using opensource-server\nCreated by: <username>"` |
| `virtualization.interface` | `name` = `eth0`, attached to the VM |
| `ipam.ip-address` | Container's IPv4 in `/32` CIDR, assigned to `eth0`, set as `primary_ip4` |

---


cmyers-mieweb:
image

image

image

create-a-container/bin/create-container.js
```diff
@@ -492,7 +481,27 @@ async function main() {
...
+      try {
+        await createVirtualMachine(baseUrl, token, {
+          hostname: container.hostname,
+          clusterName: site.name,
+          ipv4Address,
+          createdBy: container.username,
+          nodeName: container.node?.name,
+          vcpus: 4,
+          memoryMb: 4096,
+          diskGb: 50,
```
runleveldev: These resources should be based on the resource requests subsystem so they'll mirror what the container actually has.

create-a-container/utils/netbox.js
```diff
@@ -0,0 +1,292 @@
...
+
+/**
+ * Load NetBox credentials from the Settings model and invoke a callback.
+ * Returns null without calling fn if NetBox is not configured.
+ *
+ * @param {object} Setting - Sequelize Setting model
+ * @param {function(string, string): Promise<*>} fn - Called with (baseUrl, token)
+ * @returns {Promise<*|null>}
+ */
+async function withNetbox(Setting, fn) {
```
runleveldev, line 297: I'd love to know the rationale of this. I would have done a NetboxApi class with methods rather than this callback-based system. Was there a reason this was picked over an API object? If so, can we stick the reasoning in the comment here. The class-based system would have been more consistent with how we're interacting with the Proxmox API for example so I feel exceptions should be documented.

runleveldev (changes requested):

create-a-container/utils/netbox.js
cmyers-mieweb: Going to include a design note in a comment above the method call, but here is what it boils down to that I've found.
 1. Stateless auth. NetBox uses a static `Authorization: Token` header, so there is no per-session state (cookie or ticket expiry) worth holding on an instance. ProxmoxApi is a class precisely because its ticket/CSRF auth IS stateful and benefits from a long-lived object.
2. Credentials are global, not per-entity. Proxmox creds live on each Node row, so `node.api()` reads naturally as an instance method. NetBox creds are a single system wide pair in the Settings key/value table there is no domain object to anchor an `.api()` method to.

create-a-container/bin/create-container.js
cmyers-mieweb: I've modified it to pull the container details from the build
```
    const actualCores = config['cores'] != null ? parseInt(config['cores'], 10) : null;
    const actualMemoryMb = config['memory'] != null ? parseInt(config['memory'], 10) : null;
    const actualDiskGb = parseRootfsSizeGb(config['rootfs']);
```

cmyers-mieweb:
@copilot resolve the merge conflicts in this pull request

cmyers-mieweb:
@copilot resolve the merge conflicts in this pull request

runleveldev (approved):

PR #374, opened 2026-06-23
Feature: "All containers" toggle for admins sitewide; add additional docker build failsafe's. 

cmyers-mieweb:
Issue: https://github.com/mieweb/opensource-server/pull/374

This pull request introduces an "All Containers" admin view to the container management system, allowing admins to see every container on a site, including the owner of each container. It also refactors the Docker Compose setup to prevent race conditions during dependency installation, and extends the API, client types, and OpenAPI schema to support the new functionality.

**Admin "All Containers" View and API Enhancements:**

* Added a new `/sites/{siteId}/containers/all` API endpoint (admin-only) that returns all containers for a site, including the `owner` field, with support for filtering by `nodeId`, `hostname`, and `username`. [[1]](diffhunk://#diff-bdedb7b5057f4a7c805abbe4ef41948dd186af96f25737d1c806c2c97ba31151R344-R359) [[2]](diffhunk://#diff-44647bcd3b853c60c4948db454e8deec1dd1e5aa05f1dbdf4e42ef9daa2e0f22R253-R272) [[3]](diffhunk://#diff-44647bcd3b853c60c4948db454e8deec1dd1e5aa05f1dbdf4e42ef9daa2e0f22R116-R119) [[4]](diffhunk://#diff-44647bcd3b853c60c4948db454e8deec1dd1e5aa05f1dbdf4e42ef9daa2e0f22R164-R201)
* Updated the OpenAPI schema to document the new endpoint, the `owner` field on containers, and the new query parameters. [[1]](diffhunk://#diff-bdedb7b5057f4a7c805abbe4ef41948dd186af96f25737d1c806c2c97ba31151R91) [[2]](diffhunk://#diff-bdedb7b5057f4a7c805abbe4ef41948dd186af96f25737d1c806c2c97ba31151R314) [[3]](diffhunk://#diff-bdedb7b5057f4a7c805abbe4ef41948dd186af96f25737d1c806c2c97ba31151R344-R359)
* Updated the backend container serialization and query logic to include the `owner` field and to support flexible filtering for both user and admin list endpoints. [[1]](diffhunk://#diff-44647bcd3b853c60c4948db454e8deec1dd1e5aa05f1dbdf4e42ef9daa2e0f22R116-R119) [[2]](diffhunk://#diff-44647bcd3b853c60c4948db454e8deec1dd1e5aa05f1dbdf4e42ef9daa2e0f22L199-R242) [[3]](diffhunk://#diff-44647bcd3b853c60c4948db454e8deec1dd1e5aa05f1dbdf4e42ef9daa2e0f22R164-R201) [[4]](diffhunk://#diff-44647bcd3b853c60c4948db454e8deec1dd1e5aa05f1dbdf4e42ef9daa2e0f22R253-R272)

**Frontend: Containers List Page and API Integration:**

* Enhanced the `ContainersListPage` to support an "all"/"mine" toggle (visible to admins), display the container owner in the "All" view, and handle new query parameters for filtering. [[1]](diffhunk://#diff-65e8c60e95f1cfe5f50137f11eb75479913707599687553c516efc586e85120aR56) [[2]](diffhunk://#diff-8738a2103e2d5a44bda30689de14774ed7a0cb19be5bf96eae5af57b973cf5afL212-R223) [[3]](diffhunk://#diff-8738a2103e2d5a44bda30689de14774ed7a0cb19be5bf96eae5af57b973cf5afL240-R247) [[4]](diffhunk://#diff-8738a2103e2d5a44bda30689de14774ed7a0cb19be5bf96eae5af57b973cf5afL259-R287) [[5]](diffhunk://#diff-8738a2103e2d5a44bda30689de14774ed7a0cb19be5bf96eae5af57b973cf5afL331-R359) [[6]](diffhunk://#diff-8738a2103e2d5a44bda30689de14774ed7a0cb19be5bf96eae5af57b973cf5afR386-R393) [[7]](diffhunk://#diff-8738a2103e2d5a44bda30689de14774ed7a0cb19be5bf96eae5af57b973cf5afR418) [[8]](diffhunk://#diff-8738a2103e2d5a44bda30689de14774ed7a0cb19be5bf96eae5af57b973cf5afR435-R442)
* Updated client-side types and queries to handle the `owner` field and the new `listAllContainers` API. [[1]](diffhunk://#diff-36c13c4ddc6c36d2b82650e648811a2098216150795ca8a0d4973b5ea6a1716aR76) [[2]](diffhunk://#diff-3ec7271b4d8a874c93fe3877afeab9bc537844a6e1ecc0703884d0257f2e7b82R30-R31) [[3]](diffhunk://#diff-3ec7271b4d8a874c93fe3877afeab9bc537844a6e1ecc0703884d0257f2e7b82R67-R76)

**Docker Compose Dependency Installation Improvements:**

* Refactored the Docker Compose setup so only the `node` service runs `npm ci` (with `--ignore-scripts` and manual patching), preventing race conditions and ensuring reliable dependency setup for both server and client. The `client` service now waits for `node` to finish before building. [[1]](diffhunk://#diff-3493e6b5ddf34891e572f911db893efd9e46af828e011ea778a7c1eb64763588R35-R52) [[2]](diffhunk://#diff-3493e6b5ddf34891e572f911db893efd9e46af828e011ea778a7c1eb64763588L43-R62) [[3]](diffhunk://#diff-3493e6b5ddf34891e572f911db893efd9e46af828e011ea778a7c1eb64763588R82-R92)

image
image
image
image

### DIFF
--- a/compose.yml
+++ b/compose.yml
@@ -28,39 +28,15 @@ services:
           skopeo copy docker://$${IMAGE_REF} oci-archive:$${FILENAME}
         fi
 
-  # This service runs once when bringing the compose stack up to ensure the
-  # node_modules are populated in the user's create-a-container directory. This
-  # is usually handled by the manager's Dockerfile, but we're mounting over that
-  # directory so we need to be sure the user has them.
-  #
-  # This is the SINGLE owner of `npm ci` for both the server and the client.
-  # The `client` service below must NOT also run `npm ci`, otherwise two
-  # concurrent installs race on the same bind-mounted node_modules and wipe the
-  # .bin symlinks (manifesting as "tsc: not found" / "vite: not found").
-  #
-  # We install the FULL dependency tree (no --omit=dev) because the Proxmox dev
-  # systemd override runs node_modules/.bin/nodemon, and nodemon is a
-  # devDependency. The catch is another devDependency, @vscode/sqlite3 (the
-  # local-dev sqlite driver), is a native module that builds from source via
-  # node-gyp and would need Python + a C/C++ toolchain this slim image lacks.
-  # --ignore-scripts skips that build (and every other lifecycle script) so the
-  # install can't fail on it; bin symlinks like nodemon are still created
-  # because linking is not a lifecycle script. This stack runs against
-  # PostgreSQL and only require()s @vscode/sqlite3 lazily for the sqlite dialect,
-  # so leaving it unbuilt is harmless. We then run patch-package by hand to
-  # re-apply patches/ (e.g. the sequelize patch) that the skipped postinstall
-  # would have applied.
   node:
-    image: node:24-trixie-slim
+    image: node:24-trixie
     volumes:
       - ./:/opt/opensource-server
     working_dir: /opt/opensource-server/create-a-container
     entrypoint: ["/bin/sh", "-c"]
     command:
       - |
-        npm ci --ignore-scripts --no-audit --no-fund
-        ./node_modules/.bin/patch-package
-        exec npm --prefix client ci --no-audit --no-fund
+        exec npm ci --no-audit --no-fund
 
   # Watches the React client and rebuilds the production bundle on change. The
   # Manager LXC inside Proxmox serves create-a-container/client/dist statically
@@ -79,17 +55,14 @@ services:
       - ./:/opt/opensource-server
     working_dir: /opt/opensource-server/create-a-container/client
     entrypoint: ["/bin/sh", "-c"]
-    # No `npm ci` here: the `node` service is the single install owner. Running
-    # it again concurrently would race on the shared bind-mounted node_modules
-    # and remove the tsc/vite bin symlinks. We wait for `node` to finish, then
-    # build once and watch.
+    # Owns the client's node_modules install (separate directory from the
+    # server, so it runs in parallel with the `node` service). Then builds once
+    # and watches.
     command:
       - |
+        npm ci --no-audit --no-fund
         npm run build
         exec npm run build:watch
-    depends_on:
-      node:
-        condition: service_completed_successfully
     # Healthy once an initial bundle exists, so Proxmox can wait for it on a
     # fresh checkout (where client/dist isn't present yet) before serving.
     healthcheck:
--- a/compose.yml
+++ b/compose.yml
@@ -28,6 +28,10 @@ services:
           skopeo copy docker://$${IMAGE_REF} oci-archive:$${FILENAME}
         fi
 
+# This service runs once when bringing the compose stack up to ensure the
+# node_modules are populated in the user's create-a-container directory. This
+# is usually handled by the manager's Dockerfile, but we're mounting over that
+# directory so we need to be sure that the user has them.
   node:
     image: node:24-trixie
     volumes:
--- a/compose.yml
+++ b/compose.yml
@@ -32,6 +32,24 @@ services:
   # node_modules are populated in the user's create-a-container directory. This
   # is usually handled by the manager's Dockerfile, but we're mounting over that
   # directory so we need to be sure the user has them.
+  #
+  # This is the SINGLE owner of `npm ci` for both the server and the client.
+  # The `client` service below must NOT also run `npm ci`, otherwise two
+  # concurrent installs race on the same bind-mounted node_modules and wipe the
+  # .bin symlinks (manifesting as "tsc: not found" / "vite: not found").
+  #
+  # We install the FULL dependency tree (no --omit=dev) because the Proxmox dev
+  # systemd override runs node_modules/.bin/nodemon, and nodemon is a
+  # devDependency. The catch is another devDependency, @vscode/sqlite3 (the
+  # local-dev sqlite driver), is a native module that builds from source via
+  # node-gyp and would need Python + a C/C++ toolchain this slim image lacks.
+  # --ignore-scripts skips that build (and every other lifecycle script) so the
+  # install can't fail on it; bin symlinks like nodemon are still created
+  # because linking is not a lifecycle script. This stack runs against
+  # PostgreSQL and only require()s @vscode/sqlite3 lazily for the sqlite dialect,
+  # so leaving it unbuilt is harmless. We then run patch-package by hand to
+  # re-apply patches/ (e.g. the sequelize patch) that the skipped postinstall
+  # would have applied.
   node:
     image: node:24-trixie-slim
     volumes:
@@ -40,7 +58,8 @@ services:
     entrypoint: ["/bin/sh", "-c"]
     command:
       - |
-        npm ci --omit=dev --no-audit --no-fund
+        npm ci --ignore-scripts --no-audit --no-fund
+        ./node_modules/.bin/patch-package
         exec npm --prefix client ci --no-audit --no-fund
 
   # Watches the React client and rebuilds the production bundle on change. The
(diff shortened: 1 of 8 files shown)


compose.yml
```diff
@@ -40,7 +58,8 @@ services:
     entrypoint: ["/bin/sh", "-c"]
     command:
       - |
-        npm ci --omit=dev --no-audit --no-fund
+        npm ci --ignore-scripts --no-audit --no-fund
+        ./node_modules/.bin/patch-package
         exec npm --prefix client ci --no-audit --no-fund
```
runleveldev: I was starting to fix this same issue in my own branch, so a couple of notes:

* removing --omit=dev is good
* using --ignore-scripts is bad. I'd rather use an image that is able to run the scripts. `node:24-trixie` (no `-slim`) can do so. If we do this the dedicated `./node_modules/.bin/patch-package` line can be dropped.
* We should leave the client dep install in the client service and remove it from here so they can run in parallel, which will speed up the overall build time.

images/proxmox-ve/create-manager.sh
```diff
@@ -11,9 +11,22 @@ until [ -d /etc/pve/local ]; do
...
+# check on the config file is not enough: an interrupted prior `pct create`
+# (common under the nested Proxmox-in-Docker, e.g. a failed `--start=1`) leaves
+# a partial config behind that is still holding the `create` lock and missing
+# required keys like `arch`. Treating that half-created container as "done"
+# wedges every subsequent boot (`pct start` then fails with "missing 'arch'").
+# So only short-circuit when the config exists, carries no lingering create
+# lock, and has a real `arch` line; otherwise tear the remnants down and rebuild.
+CONF="/etc/pve/lxc/${CTID}.conf"
+if [ -f "${CONF}" ]; then
+    if ! grep -q '^lock: create' "${CONF}" && grep -q '^arch:' "${CONF}"; then
```
runleveldev: Guarding on `arch` seems oddly specific? Is there a noted scenario where the create lock would be gone and the arch still missing, or can we simplify this guard to just the create lock?

create-a-container/openapi.v1.yaml
```diff
@@ -339,6 +341,22 @@ paths:
       summary: Form bootstrap data (domains + NVIDIA flag)
       parameters: [{ in: path, name: siteId, required: true, schema: { type: integer } }]
       responses: { '200': { description: Bootstrap payload } }
+  /sites/{siteId}/containers/all:
```
runleveldev: I won't die on this hill, but I would have prefered if there was a way to do this in the existing /containers path, perhaps using query params. e.g. `?user=*` or something. It would slot in better with the existing hostname filter query param. We could default the frontend to querying `?user=<logged in user>` so even admins see their own containers by default, then on the back-end the query becomes (pseudocode): `user == current_user || (curent_user.is_admin && user == req.params.user)`

runleveldev (changes requested):

images/proxmox-ve/create-manager.sh
cmyers-mieweb: I've modified the conditional so the guard is now just the create-lock check

compose.yml
cmyers-mieweb: I have adjusted the image to remove the `-slim` variation of the image and removed the `--prefix-client` from the command.

create-a-container/openapi.v1.yaml
cmyers-mieweb: For the backend I've gone ahead and deleted the `/containers/all` route and dropped the `apiAdmin` import. As for the frontend, the "Mine/All" toggle now links to `/containers` and `/containers?user=*` to make the page bookmarkable.

runleveldev:
It's not a blocker to this PR, but can we see why the job for the preview container failed and get another issue opened. https://github.com/mieweb/opensource-server/actions/runs/28060284274/job/83073005593?pr=374

create-a-container/routers/api/v1/containers.js
```diff
@@ -157,6 +161,69 @@ function serializeContainer(c, site, status) {
...
+ *   - absent/empty -> the requesting user's own containers (default for all)
+ *   - '*'          -> every owner on the site (admin only)
+ *   - '<username>' -> that specific owner (admin only, unless it's yourself)
+ * Non-admins requesting anyone else (or '*') get 403 rather than a silently
+ * empty list, so the boundary is explicit.
+ * @param {string|undefined} requestedUser - req.query.user
+ * @param {{ user: string, isAdmin: boolean }} session - req.session
+ * @returns {string|undefined} username to filter by, or undefined for "all"
+ */
+function resolveUsernameFilter(requestedUser, session) {
```
runleveldev, line 217: I think this function should work differently. I'm thinking of the permissions like a venn-diagram. The two circles are "Requested User" and "Containers you have access to". For an admin (who has access to all containers) they always get the middle bit. For a non-admin, who only has access to their own containers, the venn-diagram _should_ be a circle, but if they request a user other than themselves they get nothing instead. What I'm trying to get at is I don't want this to be a hard 403 failure for someone who's non admin. Consider

|Is Admin|Requested User|Returned Containers|
|---|---|---|
|N|Own|Own|
|N|All|Own|
|N|Someone else|None|
|Y|Own|Own|
|Y|All|All|
|Y|Someone else|Theirs|

create-a-container/openapi.v1.yaml
```diff
@@ -310,7 +311,18 @@ paths:
...
-      responses: { '200': { description: Array of containers } }
+        - { in: query, name: nodeId, schema: { type: integer }, description: Filter to a single node belonging to the site }
+        - in: query
+          name: user
+          schema: { type: string }
+          description: >-
+            Owner filter. Omitted/empty returns the requesting user's own
+            containers (default). `*` returns every owner on the site and a
+            specific username returns that owner's containers; both are
+            admin-only and yield 403 for non-admins (except requesting yourself).
```
runleveldev: Per my other comment, rather than returning 403, I want the returned list to be the logical intersection of containers I have access to and containers I requested. See that comment for more details.

runleveldev (changes requested):

create-a-container/routers/api/v1/containers.js
runleveldev: The low-level idea here is to leave the door open for a more complex permission model in the future. There's a universe where we want a non-admin gets access to containers they don't own but still let them filter by username using this same mechanism.

cmyers-mieweb: `resolveUsernameFilter` now returns the requester's own username for non-admins requesting `user=*` (instead of 403). Admins still get all owners, however requesting a specific other owner still 403s.

cmyers-mieweb: once this merges I'll work on "group" functionality and container sharing between individuals and created groups

create-a-container/openapi.v1.yaml
cmyers-mieweb: see above

create-a-container/client/src/pages/containers/ContainersListPage.tsx
```diff
@@ -215,7 +216,20 @@ export function ContainersListPage() {
...
+  // `user=*` lists every owner on the site for admins; for non-admins the
+  // server scopes it back to their own containers, so the toggle is available
+  // to everyone (ahead of future shareable/collaborative containers). The param
+  // is the single source of truth so the view is bookmarkable and slots in
+  // beside the existing hostname/nodeId filters.
+  const isAll = searchParams.get('user') === '*';
+  const userFilter = isAll ? '*' : undefined;
+  // Only admins ever see more than one owner in the `*` view, so the owner
+  // column is meaningful for them alone.
+  const showOwner = isAdmin && isAll;
```
runleveldev, line 232: Not a blocker just a note, this will no longer be true once the sharing system is added.

create-a-container/routers/api/v1/containers.js
runleveldev: We're still throwing 403 here when I would have rather returned success and no data. But per your goal of container sharing, I won't consider that blocking for this PR.

runleveldev (approved):

create-a-container/client/src/pages/containers/ContainersListPage.tsx
cmyers-mieweb: I'll remove this on my next PR

PR #383, opened 2026-06-24
feat(manager): add Proxmox cluster link to Containers page

horner:
## Summary

Adds a **Proxmox** button to the Containers page header (`/sites/:siteId/containers`) that opens the Proxmox cluster web UI (e.g. https://os.mieweb.org:8006) in a new tab.

Screenshot 2026-06-24 at 7 37 54 AM

image

## Details

- The button sits between **Nodes** and **New container** in the page header.
- The cluster URL is **derived from the origin of a node's API URL** in the container list, so no cluster host is hardcoded — it adapts to any deployment.
- The button only renders when a node API URL is available (`new URL(apiUrl).origin`), and silently hides if none can be parsed.
- Opens in a new tab with `rel="noopener noreferrer"` and an accessible `aria-label` ("Open Proxmox cluster web UI").

## Testing

Verified locally via `make dev`: the button renders in the header and links to the cluster origin.

### DIFF
--- a/create-a-container/client/src/pages/containers/ContainersListPage.tsx
+++ b/create-a-container/client/src/pages/containers/ContainersListPage.tsx
@@ -253,6 +253,19 @@ export function ContainersListPage() {
 
   const hasContainers = !!data && data.length > 0;
 
+  // The Proxmox cluster web UI lives at the origin of any node's API URL
+  // (e.g. https://os.mieweb.org:8006). Derive it from the container list so we
+  // never hardcode a cluster host.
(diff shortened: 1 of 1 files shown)


runleveldev:
Clicking the node name already opens the node directly to the selected container. This feature is both redundant and less useful since it doesn't also navigate to the specific container.

horner:
Where. No one knows where?

horner:
There is no clickable way to get there from that screen.

runleveldev:
In both views, you click the node name. I'm not against making this more prominent, but it is here.
image

PR #388, opened 2026-06-30
Add container sharing (collaborators) feature

### LINKED ISSUE #389: [Bug]: User filter doesn't apply to container listing
### Current URL

/sites/1/containers?user=REDACTED

### Username

rgingras

### Version

v2026.6.5

### Steps to reproduce

1. Go to /sites/1/containers
2. Select "all"
3. Change ?user=* in the url to ?user=\<anyone else>
4. See your own containers

### Expected behavior

Instead of seeing your own containers, you should see the user in the URL parameter.

### Actual behavior

Setting the URL parameter to anything other than * results in seeing your own containers.

### Additional context

_No response_

### LINKED ISSUE #390: Make the "all containers" screen the default/only screen with container filters
Instead of seperate "Mine" and "All" containers screens, use only one containers screen with a filters system. The filters stored either in the URL or LocalStorage and default to the filter `user == me` if none are selected.

Image

### LINKED ISSUE #391: [Bug]: 404 for non-owned container as Admin
### Current URL

/api/v1/sites/1/containers/56

### Username

rgingras

### Version

v2026.6.5

### Steps to reproduce

1. View all containers
2. Select "Edit" on a container you don't own
3. View empty form. Network tab shows 404 on the /api/v1/sites/.../:id request

### Expected behavior

Admins should be able to manage containers they don't own.

### Actual behavior

Admins can view in the container list, but not get details of or modify, containers they don't own.

### Additional context

_No response_

cmyers-mieweb:
Introduce container sharing/collaborators so containers can be shared with additional users.

Details:
- Backend: add ContainerCollaborator model and migration with unique (containerId, username) constraint and CASCADE deletion.
- DB/model changes: Container.hasMany(ContainerCollaborator) association and eager-load collaborators in container queries/serializer.
- Permissions: implement userCanAccess / userCanManage helpers and loadContainerForSession to authorize view vs manage actions.
- Listing/filtering: applyOwnershipFilter folds shared containers into the "All" view for non-admins and preserves admin behavior for '*'.
- Share API: GET/POST/DELETE /sites/:siteId/containers/:id/collaborators to list, add, and remove collaborators; share/unshare endpoints validate users and return the updated collaborator list.
- Creation: Container create accepts additionalOwners array; creator-specified shares validated and stored in the same transaction.
- Deletion: container deletion removes collaborator grants explicitly and enforces owner/admin-only deletion.
- Client: add queries.shareContainer/unshareContainer, Container.collaborators type, and React components (CollaboratorChips, AddCollaboratorField, CollaboratorsManager); integrate into ContainerFormPage (create/edit) and ContainersListPage (share modal and row action).

This change enables collaborative workflows while enforcing owner/admin management and consistent UI feedback.

image
image
image
image
image
image


cmyers-mieweb:
Modifying "all" container toggle with a more clear and defined filter system to resolve https://github.com/mieweb/opensource-server/issues/390 

Also additional bug fixes to resolve
https://github.com/mieweb/opensource-server/issues/389
https://github.com/mieweb/opensource-server/issues/391

image
image
image

cmyers-mieweb:
Also added color over hovered buttons for better UI 
image

create-a-container/openapi.v1.yaml
```diff
@@ -339,6 +339,10 @@ paths:
                 customTemplate: { type: string }
                 entrypoint: { type: string }
                 nvidiaRequested: { type: boolean }
+                additionalOwners:
```
runleveldev: I don't like the wording of "additional owners". Collaborators is much better because it doesn't indicate a higher level of permissions than we're actually granting. (If we do enable real extra owners in future "co-owners" would be the correct term anyways).

create-a-container/routers/api/v1/containers.js
```diff
@@ -200,30 +209,162 @@ function buildContainerListWhere(query, nodeIds) {
...
+ * @returns {string[]}
+ */
+function parseUserFilter(raw) {
+  if (raw === undefined || raw === null) return [];
+  const parts = (Array.isArray(raw) ? raw : String(raw).split(','))
+    .map((s) => String(s).trim())
+    .filter(Boolean);
+  return [...new Set(parts)];
+}
+
```
runleveldev: This is overkill. We should rely on express's query string parsing behavior to properly implement the parsing into a list. By setting express's query parser to `extended` (https://expressjs.com/en/5x/api/application/#application-settings) we can ensure the `user` parameter is always a list since it accepts the form of `user[0]=value` to coerce into a list even on single values. Routes that use this function only then need a single line: `req.query.user ??= []` to set the query to the empty list if it's otherwise undefined.

create-a-container/routers/api/v1/containers.js
```diff
@@ -200,30 +209,162 @@ function buildContainerListWhere(query, nodeIds) {
...
+ * @returns {Promise<object[]>}
+ */
+async function ownVisibleClauses(username) {
+  const shares = await ContainerCollaborator.findAll({
+    where: { username },
+    attributes: ['containerId'],
+  });
+  const sharedIds = shares.map((s) => s.containerId);
+  return [{ username }, ...(sharedIds.length > 0 ? [{ id: sharedIds }] : [])];
+}
```
runleveldev: This function runs a query to get it's information before sticking that into another query later on to find what we're actually looking for. If I'm reading it correctly, this should just be a JOIN to the "containers" model which should be more efficient since it's one less round trip to the database.

create-a-container/routers/api/v1/containers.js
```diff
@@ -200,30 +209,162 @@ function buildContainerListWhere(query, nodeIds) {
...
+ * Whether a session may manage a container's sharing (add/remove collaborators)
+ * or delete it: only its primary owner or an admin. Collaborators can use a
+ * shared container but cannot re-share or delete it.
+ * @param {object} container - Container instance.
+ * @param {object} session - req.session ({ user, isAdmin }).
+ * @returns {boolean}
+ */
+function userCanManage(container, session) {
+  return session.isAdmin || container.username === session.user;
+}
```
runleveldev: These permission checks should be methods on the container model. Container#canView(username) checks that that username is allowed to view the container and Container#canEdit(username) check that that username is allowed to edit it. Those methods can encapsulate more complex logic down the line.

create-a-container/routers/api/v1/containers.js
```diff
@@ -200,30 +209,162 @@ function buildContainerListWhere(query, nodeIds) {
...
+ * leaks the existence of containers the caller may not see.
+ * @param {object} req - Express request.
+ * @param {object} [opts]
+ * @param {boolean} [opts.requireManage=false] - Require owner/admin (sharing,
+ *   delete) rather than the looser view/edit access.
+ * @param {Array} [opts.include] - Override the eager-load graph.
+ * @returns {Promise<{site: object, container: object}>}
+ */
+async function loadContainerForSession(req, { requireManage = false, include } = {}) {
+  const site = await loadSite(req);
```
runleveldev: I know this wasn't written by this PR, by why are we passing the whole request into a function that just needs the ID?

create-a-container/routers/api/v1/containers.js
```diff
@@ -200,30 +209,162 @@ function buildContainerListWhere(query, nodeIds) {
...
+  const id = parseInt(req.params.id, 10);
+  if (!Number.isInteger(id) || id <= 0) {
+    throw new ApiError(404, 'not_found', 'Container not found');
+  }
```
runleveldev, line 294: 404 is wrong here. It's not that nothing was found, its that the data we were given didn't make sense to try to find anything with. A 400 error is correct here.

create-a-container/routers/api/v1/containers.js
```diff
@@ -200,30 +209,162 @@ function buildContainerListWhere(query, nodeIds) {
...
+  const container = await Container.findByPk(id, {
+    include: include || [{ association: 'node' }, { association: 'collaborators' }],
+  });
+  if (!container || !container.node || container.node.siteId !== site.id) {
+    throw new ApiError(404, 'not_found', 'Container not found');
+  }
+  const authorized = requireManage
+    ? userCanManage(container, req.session)
+    : userCanAccess(container, req.session);
+  if (!authorized) throw new ApiError(404, 'not_found', 'Container not found');
```
runleveldev: This should all be a single query, not post-query server logic. Consider:
1. If manage is required
    2. If user is admin: no user filter on the query.
    3. Otherwise: only select containers that the current user is the owner of
2. If manage is not required: join to collaborators table and filter on `user is owner || user is a collaborator)`

create-a-container/routers/api/v1/containers.js
```diff
@@ -200,30 +209,162 @@ function buildContainerListWhere(query, nodeIds) {
...
+ * session against it. Returns 404 (not 403) on any failure so the route never
+ * leaks the existence of containers the caller may not see.
+ * @param {object} req - Express request.
+ * @param {object} [opts]
+ * @param {boolean} [opts.requireManage=false] - Require owner/admin (sharing,
+ *   delete) rather than the looser view/edit access.
+ * @param {Array} [opts.include] - Override the eager-load graph.
+ * @returns {Promise<{site: object, container: object}>}
+ */
+async function loadContainerForSession(req, { requireManage = false, include } = {}) {
```
runleveldev: This shouldn't take the whole request, it's bad encapsulation. The function signature should be more like: `fn(siteId, containerId, username(or session object), {options})`

create-a-container/routers/api/v1/containers.js
```diff
@@ -266,12 +407,11 @@ router.get(
...
-    // sees their own containers by default. `user=*` lists every owner for admins
-    // and stays scoped to the requester for non-admins; `user=<name>` targets a
-    // specific owner (admin-only). undefined means "all owners".
-    const username = resolveUsernameFilter(req.query.user, req.session);
-    if (username !== undefined) where.username = username;
+    // `user` filter drives the Mine/All toggle and folds in shared containers:
+    // defaults to the requesting user (Mine); admins may pass `user=*` for every
+    // owner or `user=<name>` for a specific one; non-admins passing `user=*` get
+    // their own plus any containers shared with them. See applyOwnershipFilter.
+    await applyOwnershipFilter(where, req);
```
runleveldev: Don't ever rely on mutating variables passed into functios like `where` in this line. Also we already have a function that builds the clause based on the query. Instead of running this seperately, it should be encapsulated into the `buildContainerListWhere` function.

create-a-container/routers/api/v1/containers.js
```diff
@@ -296,10 +436,10 @@ router.get(
       throw new ApiError(404, 'not_found', 'Container not found');
     }
     const c = await Container.findOne({
-      where: { id: containerId, username: req.session.user },
+      where: { id: containerId },
       include: CONTAINER_INCLUDE,
     });
-    if (!c || !c.node || c.node.siteId !== site.id) {
+    if (!c || !c.node || c.node.siteId !== site.id || !userCanAccess(c, req.session)) {
```
runleveldev: As I mentioned in another comment, I don't want this access logic living in the server code, I want it as part of the queries. If we design this right the server won't see a valid container and then decide the user doesn't have the rights, if the user doesn't have the rights it'll never see the container in the first place.

create-a-container/routers/api/v1/containers.js
```diff
@@ -385,6 +526,25 @@ router.post(
...
+          const username = await resolveShareUsername(raw, container, { transaction: t });
+          if (seen.has(username)) continue;
+          seen.add(username);
+          await ContainerCollaborator.create(
+            { containerId: container.id, username },
+            { transaction: t },
+          );
+        }
+      }
+
```
runleveldev, line 503: This whole section could be a single bulkInsert into the database, using `On Conflict replace` clauses to handle deduplication and database constraints to handle unknown users. I believe that would also render the `resolveShareUsername` function obselete.

create-a-container/routers/api/v1/containers.js
```diff
@@ -463,10 +623,15 @@ router.put(
...
-      where: { id: parseInt(req.params.id, 10), username: req.session.user },
-      include: [{ model: Node, as: 'node', where: { siteId: site.id } }],
+      where: { id: parseInt(req.params.id, 10) },
+      include: [
+        { model: Node, as: 'node', where: { siteId: site.id } },
+        { association: 'collaborators' },
+      ],
     });
-    if (!container) throw new ApiError(404, 'not_found', 'Container not found');
+    if (!container || !userCanAccess(container, req.session)) {
```
runleveldev: Just pointing this out again. We're doing application-level access logic here when it could be database-level joins to ensure the server never sees a container the current user doesn't have access to.

create-a-container/routers/api/v1/containers.js
```diff
@@ -664,6 +835,9 @@ router.delete(
         console.log(`Node-side deletion skipped or failed: ${err.message}`);
       }
     }
+    // Remove sharing grants explicitly so the rows are gone regardless of
+    // whether the DB enforces the ON DELETE CASCADE foreign key.
+    await ContainerCollaborator.destroy({ where: { containerId: container.id } });
```
runleveldev: Terrible. We need to trust the database to enforce on delete cascade.

create-a-container/routers/api/v1/containers.js
```diff
@@ -675,4 +849,44 @@ router.delete(
   }),
 );
 
+// GET /containers/:id/collaborators — list users a container is shared with.
+// Visible to anyone who can access the container (owner, collaborator, admin).
+router.get(
+  '/:id/collaborators',
+  asyncHandler(async (req, res) => {
+    const { container } = await loadContainerForSession(req);
+    return ok(res, { collaborators: await listCollaboratorUsernames(container.id) });
```
runleveldev: Why are we loading the container (1 DB call) then listing the containers based on that object? Just load the ContainerCollaborator model directly with a join to the container object to properly enforce permissions.

create-a-container/routers/api/v1/containers.js
```diff
@@ -675,4 +849,44 @@ router.delete(
...
+// POST /containers/:id/collaborators — share with another user (owner/admin).
+// Body: { username }. 404 user_not_found if the username doesn't exist.
+router.post(
+  '/:id/collaborators',
+  asyncHandler(async (req, res) => {
+    const { container } = await loadContainerForSession(req, { requireManage: true });
+    const username = await resolveShareUsername(req.body?.username, container);
+    const [, isNew] = await ContainerCollaborator.findOrCreate({
+      where: { containerId: container.id, username },
+    });
```
runleveldev: This one is a little more complex than my previous comments, but the essential principle holds. The initial loadContainerForSession is fine since we need the permission check, but the "preflight" query in resolveShareUsername is redundant since the database should enforce foreign key constraints on the insert into ContainerCollaborator.

create-a-container/routers/api/v1/containers.js
```diff
@@ -675,4 +849,44 @@ router.delete(
...
+    if (!isNew) {
+      throw new ApiError(409, 'already_shared', `Already shared with ${username}`);
```
runleveldev: Since we're doing a findOrCreate I don't see why we should error here. This handler appears to me to be idempotent (creates the association only if it doesn't already exist), so there's no reason to complain that it got "double shared".

create-a-container/models/container-collaborator.js
```diff
@@ -0,0 +1,44 @@
...
+    }
+  }
+  ContainerCollaborator.init(
+    {
+      containerId: {
+        type: DataTypes.INTEGER,
+        allowNull: false,
+        references: { model: 'Containers', key: 'id' },
+      },
+      username: {
```
runleveldev, line 26: This needs to have a foreign key constraint matching Users.uid so the database can enforce existance checks. I would do an ON DELETE CASCADE here as well so a user deleted removes their collaborator entries.

create-a-container/client/src/pages/containers/ContainersListPage.tsx
```diff
@@ -163,23 +171,49 @@ function RowActions({
...
 }) {
   return (
     <>
       {c.creationJobId && (
         <Link to={`/jobs/${c.creationJobId}`}>
-          <Button variant="ghost" size="sm">
+          <Button
+            variant="ghost"
+            size="sm"
+            className="transition-colors hover:bg-slate-100 hover:text-slate-700 dark:hover:bg-slate-800 dark:hover:text-slate-200"
```
runleveldev: The Button component from `@mieweb/ui` we're using is supposed to do this automatically. Can we figure out why that's not working instead of trying to work around it?

create-a-container/client/src/pages/containers/ContainersListPage.tsx
runleveldev: I know this isn't this PRs fault but we need to fix it somewhere. These top-level component functions need to be broken out into dedicated component files under `client/src/components` rather than living alongside the page.

create-a-container/client/src/pages/containers/ContainersListPage.tsx
```diff
@@ -265,33 +359,76 @@ export function ContainersListPage() {
...
+  }, [isAdmin, allUsers, visibleContainers, sessionUser]);
+
+  const statusOptions = useMemo<FilterOption[]>(
+    () =>
+      (Object.keys(STATUS_LABELS) as ContainerStatus[]).map((s) => ({
+        value: s,
+        label: STATUS_LABELS[s],
+      })),
+    [],
+  );
```
runleveldev: This is probably fine, but we should probably derive this from the actual statuses from the API to keep the dropdown clean.

create-a-container/client/src/components/containers/ContainerFilters.tsx
```diff
@@ -0,0 +1,208 @@
...
+export interface FilterOption {
+  value: string;
+  label: string;
+}
+
+/**
+ * A compact checkbox-popover multiselect. Selected values are controlled by the
+ * parent so filter state stays in the URL. Closes on outside click.
+ */
+function MultiSelect({
```
runleveldev: `@mieweb/ui` already has a MutliSelect https://ui.mieweb.org/?path=/story/components-forms-inputs-dropdown--multi-select&globals=theme:dark

runleveldev: This has a "select all" option which I would prefer to the custom "Everone" option we have.

create-a-container/client/src/pages/containers/CollaboratorsManager.tsx
```diff
@@ -0,0 +1,165 @@
...
+            <button
+              type="button"
+              onClick={() => onRemove(username)}
+              disabled={disabled}
+              aria-label={`Remove ${username}`}
+              className="inline-flex items-center justify-center rounded-full p-0.5 hover:bg-black/10 disabled:opacity-50"
+            >
+              <X className="size-3" aria-hidden="true" />
+            </button>
+          </Badge>
```
runleveldev: I think you can use https://ui.mieweb.org/?path=/docs/components-status-indicators-servicebadge--docs&globals=theme:dark which is a Badge that comes with the X to remove.

create-a-container/client/src/pages/containers/ContainerFormPage.tsx
```diff
@@ -608,6 +624,52 @@ export function ContainerFormPage() {
           </CardContent>
         </Card>
 
+        <Card padding="none" className={sectionCardClass}>
```
runleveldev: New sections should go at the bottom of the form.

runleveldev (changes requested):
Regarding my large number of comments in `create-a-container/routers/api/v1/containers.js`:

runleveldev (changes requested):
This is happening when loading the containers list page. If I get my way I really want the behavior of /api/v1/sites/:id/containers to be equivelent to the `?user=*` form (making `user=*` redundant and obselete and hopefully removing the need for the `*` special handling in the backend).

Image

create-a-container/openapi.v1.yaml
cmyers-mieweb: Refactored to `Collaborators`

create-a-container/routers/api/v1/containers.js
cmyers-mieweb: I've simplified this a bit by deleting the parse user filter. As such `listContainers` now takes `user?:string[]` and returns `[user[0]=...&user[1]=...` so the server always receives an array.

create-a-container/routers/api/v1/containers.js
cmyers-mieweb: I've attempted to make this function more effecient by having `OwnVisibleClauses` no longer run its own query and instead the shared container check is now an `[id IN (SELECT containerId FROM ContainerCollaborators WHERE username = …`  so visibility resolves in one trip 

```
function ownVisibleClauses(username) {
  // Built with the dialect's query generator so identifier quoting and value
  // escaping stay correct across sqlite/mysql/postgres. selectQuery emits a
  // trailing ';' which is invalid inside IN (…), hence the slice.
  const shared = sequelize.dialect.queryGenerator
    .selectQuery(
      ContainerCollaborator.getTableName(),
      { attributes: ['containerId'], where: { username } },
      ContainerCollaborator,
    )
    .slice(0, -1);
  return [{ username }, { id: { [Sequelize.Op.in]: Sequelize.literal(`(${shared})`) } }];
}
```

create-a-container/routers/api/v1/containers.js
cmyers-mieweb: How's this instead
```
  const authorized =
    req.session.isAdmin ||
    (requireManage
      ? container.canEdit(req.session.user)
      : container.canView(req.session.user));
      
```

create-a-container/routers/api/v1/containers.js
cmyers-mieweb: I asked AI to review and there wasn't a good reason aside from convenience that hid the function's real dependency. `loadSite` now takes the id directly.

create-a-container/client/src/pages/containers/ContainersListPage.tsx
cmyers-mieweb: I've divided all of these into separate components.

create-a-container/openapi.v1.yaml
runleveldev: Confirmed. No references to additional owners found.

create-a-container/routers/api/v1/containers.js
runleveldev: This might not have been clear because I left so many comments. I wanted the logic in `ownVisibleClauses` and `applyOwnershipFilter` (which is the sole caller of `ownVisibleClauses`) to be folded into `buildContainerListWhere` so that it becomes the sole function arbiting the "where" clause for container queries.

create-a-container/routers/api/v1/containers.js
runleveldev: I'm resolving this despite it not being addressed since my comment on `ownVisibleClauses` will cover it.

create-a-container/client/src/pages/containers/ContainersListPage.tsx
runleveldev: Leaving a new review comment about this so It's associated with the right section.

create-a-container/client/src/components/containers/RowActions.tsx
```diff
@@ -0,0 +1,73 @@
...
+      {c.creationJobId && (
+        <Link to={`/jobs/${c.creationJobId}`}>
+          <Button
+            variant="ghost"
+            size="sm"
+            className="transition-colors hover:bg-slate-100 hover:text-slate-700 dark:hover:bg-slate-800 dark:hover:text-slate-200"
+          >
+            Logs
+          </Button>
+        </Link>
```
runleveldev: I think the builtin Button fade is being overridden by the Link wrapper. Can we fix that instead of coding our own transition?

runleveldev (changes requested):

create-a-container/client/src/components/containers/RowActions.tsx
cmyers-mieweb: I've fixed where it was being overridden, however I kept similar logic for the delete button so that it would have a distinct color when hovered over.

create-a-container/routers/api/v1/containers.js
cmyers-mieweb: I've consolidated the logic behind `applyOwnershipFilter` and `ownVisibleClauses` into `buildContainerListWhere` so it becomes a single function that accepts the session along with the query and Node ID's.

create-a-container/routers/api/v1/containers.js
runleveldev: Confirmed addressed as of ed4c275f: `Container#canView` / `Container#canEdit` are model methods now (models/container.js:37,49), the standalone `userCanAccess`/`userCanManage` helpers are gone, and `loadContainerForSession` + the detail routes all use them. Resolving.

create-a-container/routers/api/v1/containers.js
```diff
@@ -296,10 +395,15 @@ router.get(
...
     });
-    if (!c || !c.node || c.node.siteId !== site.id) {
+    if (
+      !c ||
+      !c.node ||
+      c.node.siteId !== site.id ||
+      !(req.session.isAdmin || c.canView(req.session.user))
+    ) {
       throw new ApiError(404, 'not_found', 'Container not found');
     }
```
runleveldev: This fetch+auth block (together with the `loadSite` call and id guard above) re-implements `loadContainerForSession`, which this PR adds but only uses for the collaborator routes. It can collapse to:

```js
const { site, container: c } = await loadContainerForSession(req, { include: CONTAINER_INCLUDE });
```

`CONTAINER_INCLUDE` already eager-loads `node` + `collaborators`, which is everything the helper's checks and `canView` need.

create-a-container/routers/api/v1/containers.js
```diff
@@ -461,12 +585,17 @@ router.post(
...
+      where: { id: parseInt(req.params.id, 10) },
+      include: [
+        { model: Node, as: 'node', where: { siteId: site.id } },
+        { association: 'collaborators' },
+      ],
     });
-    if (!container) throw new ApiError(404, 'not_found', 'Container not found');
+    if (!container || !(req.session.isAdmin || container.canView(req.session.user))) {
+      throw new ApiError(404, 'not_found', 'Container not found');
+    }
```
runleveldev: Same duplication here — this is exactly `loadContainerForSession`'s default behavior (`[node, collaborators]` include, `canView` rule):

```js
const { site, container } = await loadContainerForSession(req);
```

Bonus: the helper validates the id, which this route doesn't — `PUT .../containers/abc` currently reaches the DB as `id: NaN`.

create-a-container/routers/api/v1/containers.js
```diff
@@ -632,6 +762,11 @@ router.delete(
     if (!container || !container.node || container.node.siteId !== site.id) {
       throw new ApiError(404, 'not_found', 'Container not found');
     }
+    // Deleting is owner/admin only; collaborators may use but not destroy a
+    // shared container.
+    if (!(req.session.isAdmin || container.canEdit(req.session.user))) {
+      throw new ApiError(404, 'not_found', 'Container not found');
+    }
```
runleveldev: Third copy — here it's the helper with `requireManage: true` (same `canEdit` rule as this check):

```js
const { site, container } = await loadContainerForSession(req, {
  requireManage: true,
  include: [
    { association: 'node' },
    { association: 'collaborators' },
    /* existing services -> httpService -> externalDomain graph */
  ],
});
```

Replaces the fetch and both 404 checks, and adds the id validation this route is also missing.

create-a-container/routers/api/v1/containers.js
```diff
@@ -196,34 +224,106 @@ function buildContainerListWhere(query, nodeIds) {
...
+  });
+  if (!container || !container.node || container.node.siteId !== site.id) {
+    throw new ApiError(404, 'not_found', 'Container not found');
+  }
+  const authorized =
+    req.session.isAdmin ||
+    (requireManage
+      ? container.canEdit(req.session.user)
+      : container.canView(req.session.user));
+  if (!authorized) throw new ApiError(404, 'not_found', 'Container not found');
```
runleveldev: Consolidating my three resolved "auth should live in the query" threads into one actionable comment, with a refinement:

**View access belongs in this query.** If the caller can't at least view the container, the row should never come back — the query itself enforces visibility and we 404 without loading anything:

```js
const where = { id };
if (!req.session.isAdmin) {
  // owner OR shared-with-me — the same visibility rule buildContainerListWhere builds
  where[Sequelize.Op.or] = visibleToClauses(req.session.user);
}
const container = await Container.findOne({ where, include: include || [...] });
if (!container || !container.node || container.node.siteId !== site.id) {
  throw new ApiError(404, 'not_found', 'Container not found');
}
if (requireManage && !(req.session.isAdmin || container.canEdit(req.session.user))) {
  throw new ApiError(403, 'forbidden', 'Only the owner may manage this container');
}
```

**Keep the `canEdit` check app-level, but make it 403, not 404.** A caller who reached this point can already view the container, so a 403 leaks nothing new — it just tells a collaborator why they can't delete/share. (Update the docstring above, which currently promises 404-on-everything.)

Two notes:
- I know I asked you to fold `ownVisibleClauses` into `buildContainerListWhere` — now that there's a second consumer, pull just the clause-building fragment (no query, it only returns the `[{ username }, { id: IN (...) }]` array) back out into a small shared function so the owner-or-shared rule lives in exactly one place.
- This lands fully once GET/PUT/DELETE `/:id` adopt this helper, per my other comments in this review.

runleveldev (changes requested):
Requesting changes on two connected themes, both aimed at shrinking this PR and enforcing authorization in one place:

1. **Adopt your own helper.** `loadContainerForSession` (added by this PR) is re-implemented inline in `GET/PUT/DELETE /:id`. Converting those three routes to call it trims ~40 lines, keeps the access rules in a single spot, and fixes the missing id validation in PUT/DELETE (`id: NaN` currently reaches the DB).
2. **Move view-authorization into the query.** The container-loading query should return a container if and only if the caller may at least view it — no loading first and checking after (404 when invisible). Keep the `canEdit` check app-level so a viewer attempting an owner-only action gets a **403** instead of 404; that leaks nothing since they can already see the container. Details in the comment on `loadContainerForSession`, which consolidates my three older threads on this (now resolved as superseded).

Out of scope for this PR: the near-identical service-creation blocks in POST and PUT predate this branch — please leave that for a follow-up rather than growing this diff.

---

*Disclosure: this review was AI-assisted (OpenCode). My prompts, in brief: map which routes in `containers.js` call which helper functions and identify routes re-implementing helper logic; resolve my earlier review threads that the newer commits had addressed or that this review supersedes; and draft the line-anchored comments here — the adopt-the-helper suggestions and the query-level-visibility + 403-on-edit design. I reviewed the findings and comments before submitting.*

cmyers-mieweb:
I removed the model-level `canView` helper and introduced `visibleToClauses` to centralize the owner-or-shared visibility rule as Sequelize WHERE clauses. Now we use this helper in list and single-container loaders so view access is enforced in DB queries (non-admins only load containers they own or that are shared with them). Updated `loadContainerForSession` to support `requireManage` which returns 403 for manage-only operations (owner/admin), and adjust routes (GET/PUT/DELETE and collaborators endpoints) to use the new loader and stricter checks. Also updated OpenAPI docs to document 403 responses for delete/share/unshare. Misc: simplify query-building by replacing inline subquery logic with `visibleToClauses `and tidy eager-loads.

create-a-container/routers/api/v1/containers.js
cmyers-mieweb: ` loadContainerForSession` now throws 400 `invalid_request` for non-numeric/non-positive ids

create-a-container/routers/api/v1/containers.js
cmyers-mieweb: signature is now `loadContainerForSession(siteId, containerId, session, {options});` all five call sites updated.

create-a-container/routers/api/v1/containers.js
cmyers-mieweb: Replaced the loop with one [ContainerCollaborator.bulkCreate](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/4fe60c8b1c/resources/app/out/vs/code/electron-browser/workbench/workbench.html) ([ignoreDuplicates](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/4fe60c8b1c/resources/app/out/vs/code/electron-browser/workbench/workbench.html) → ON CONFLICT DO NOTHING) that relies on the new [Users.uid](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/4fe60c8b1c/resources/app/out/vs/code/electron-browser/workbench/workbench.html) FK for existence, making resolveShareUsername obsolete.

create-a-container/routers/api/v1/containers.js
cmyers-mieweb: Removed the explicit [destroy](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/4fe60c8b1c/resources/app/out/vs/code/electron-browser/workbench/workbench.html),  container deletion now relies solely on the [containerId](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/4fe60c8b1c/resources/app/out/vs/code/electron-browser/workbench/workbench.html)

create-a-container/routers/api/v1/containers.js
cmyers-mieweb: `GET /:id/collaborators` now serializes from the collaborators already eager-loaded by the permission-checking query, so there's no second query.

create-a-container/routers/api/v1/containers.js
cmyers-mieweb: Removed the already_shared 409, sharing with an existing collaborator is now an idempotent no-op that returns the current list.

create-a-container/models/container-collaborator.js
cmyers-mieweb: Added references: `Users.uid` with ON DELETE/UPDATE CASCADE to the migration and model, so the DB enforces existence and deleting a user removes their grants.

create-a-container/routers/api/v1/containers.js
cmyers-mieweb: Implemented as suggested: `visibleToClauses` is the shared owner-or-shared fragment, view access is enforced in the loader's query, and the app-level `canEdit` check returns 403 with the docstring updated.

runleveldev (approved):

PR #392, opened 2026-07-07
Fix #307: pin sshd LoginGraceTime to cover the 2FA push window

### LINKED ISSUE #307: [Bug]: SSH Timeout too short
### Current URL

_No response_

### Username

_No response_

### Version

_No response_

### Steps to reproduce

1. SSH to a container
2. SSH timeout is shorter than 2FA timeout

### Expected behavior

SSH Timeout should be 60 seconds.

### Actual behavior

SSH Timeout should be 60 seconds.

### Additional context

_No response_

Om-singhaI:
## Issue

Fixes #307 — *SSH Timeout too short.* The SSH login window can be shorter than the 2FA timeout, so a user who answers the push in time still gets disconnected.

## Root cause (my reading — please confirm)

SSH logins into a container authenticate through `sssd` → the LDAP gateway, which fires a push‑notification 2FA. The sssd side already allows up to **60s** for the push to be answered:

```
# images/base/sssd.conf.template
# set a timeout long enough for a push notification to be responded to
ldap_opt_timeout = 60
```

But the container's sshd has no matching `LoginGraceTime`, so sshd's default grace window can drop the connection before that 60s push window closes.

## Change

Add a dedicated `sshd_config.d` drop-in setting `LoginGraceTime 75` and copy it into the base image. 75s sits just above the 60s 2FA timeout, leaving headroom for the TCP + key-exchange handshake that precedes the LDAP bind.

- `images/base/50-login-grace-time.conf` (new)
- `images/base/Dockerfile` — copy the drop-in into `/etc/ssh/sshd_config.d/`

Follows the existing drop-in pattern (`50-sss-ssh-authorizedkeys.conf`) and keeps the change config-only.

## Notes / open question

The issue is terse, so I inferred the mechanism from the code rather than a live repro. If the intended value is exactly **60s** (matching the issue text) rather than 75s with headroom, I'm happy to change it. And if the real timeout lives somewhere other than the container's sshd (e.g. a gateway/proxy in front), point me at it and I'll move the fix.

### DIFF
--- a/images/base/50-login-grace-time.conf
+++ b/images/base/50-login-grace-time.conf
@@ -1,0 +1,7 @@
+# Give SSH logins enough time to answer the push-notification 2FA prompt.
+# Authentication flows through sssd -> the LDAP gateway, which allows up to
+# 60s for a push to be responded to (see ldap_opt_timeout in sssd.conf.template).
+# sshd's default LoginGraceTime can cut the connection before that window
+# closes, so pin it above the 60s 2FA timeout, leaving headroom for the TCP
+# and key-exchange handshake that precedes the LDAP bind. See issue #307.
+LoginGraceTime 75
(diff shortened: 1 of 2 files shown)


runleveldev:
The default LoginGraceTime setting in OpenSSH is 120s so I don't agree that this is the root cause.

```
     LoginGraceTime
             The server disconnects after this time if the user has not successfully logged in.  If the value is 0, there is no time limit.  The default is 120 seconds.
```

(from `man sshd_config`)

runleveldev:
@Om-singhaI we've determined the root cause of this issue is with VSCode, not with SSH on the individual containers. Thank you for the contribution but it is not nessecary at this time.

PR #394, opened 2026-07-08
feat(nodes): add Docker API backend

### LINKED ISSUE #375: Docker-type Node API
Using the generic API contract implemented by the DummyAPI and ProxmoxAPI classes, implement a new DockerAPI class which is able to interact with a local or remote docker socket (same format as the DOCKER_HOST environment variable) via it's native HTTP(S) API (i.e. not shelling-out to the docker CLI). Implement any changes to the NodeAPI abstraction as needed to ensure the abstraction holds (i.e. no `if (nodeType == 'docker')` or similar statements outside of the initial NodeAPI dispatch).

ArshSSandhu:
## Summary

Closes #375.

This PR adds support for Docker-backed nodes by introducing a new `DockerApi` implementation that follows the existing Node API contract used by `ProxmoxApi` and `DummyApi`.

Changes include:

- Added `DockerApi` using Docker Engine's native HTTP API through `axios`
- Added support for Docker host formats:
  - `unix:///var/run/docker.sock`
  - `tcp://host:2375`
  - `http://...`
  - `https://...`
- Updated `Node.api()` to dispatch to `DockerApi` when `nodeType === "docker"`
- Added provider-neutral API access checks through `node.hasApiAccess()`
- Updated container provisioning node selection to use `Node.provisionableWhere(...)`
- Added `nodeType` support to node API serialization, create, and update routes
- Updated the node form UI to support Proxmox, Docker, and dummy nodes
- Allowed Docker host values such as `unix:///var/run/docker.sock` in the frontend form

## Related Issue

Closes #375

## Testing

Tested locally with the following checks:

- Ran backend syntax checks successfully:

  ```bash
  node --check create-a-container/models/node.js
  node --check create-a-container/utils/docker-api.js
  node --check create-a-container/utils/container-status.js
  node --check create-a-container/routers/api/v1/containers.js
  node --check create-a-container/routers/api/v1/nodes.js
  ```

- Ran frontend type-check successfully:

  ```bash
  npm run client:type-check
  ```

- Ran whitespace checks successfully:

  ```bash
  git diff --check
  git diff --cached --check
  ```

- Tested `DockerApi` directly against local Docker using:

  ```txt
  unix:///var/run/docker.sock
  ```

- Verified direct Docker lifecycle flow:
  - Pulled `nginx:latest`
  - Created a Docker container
  - Started the container
  - Checked running status
  - Deleted the test container

- Verified the UI can create a Docker node with:

  ```txt
  Name: local-docker
  Node type: Docker
  Docker host: unix:///var/run/docker.sock
  ```

- Attempted full UI container creation with `nginx:latest`. The job selected the Docker node and Docker image path correctly, but failed with:

  ```txt
  connect ENOENT /var/run/docker.sock
  ```

  This appears to be a local runtime/socket-mount limitation because the Manager runtime does not currently have the host Docker socket mounted. The `DockerApi` itself was verified successfully through the direct lifecycle test above.

### Docker node form

image

### Docker node created

image

## Breaking Changes

None known.

## Notes

Full end-to-end Docker provisioning from the UI requires the Manager runtime to have access to the configured Docker host. For local socket usage, `/var/run/docker.sock` must be mounted or otherwise reachable from the process running Manager.


create-a-container/client/src/pages/nodes/NodeFormPage.tsx
```diff
@@ -12,23 +12,24 @@ import {
...
-import { api, ApiError } from '@/lib/api';
-import { keys, queries } from '@/lib/queries';
-import { FormPageLayout } from '@/components/FormPageLayout';
-import type { Node } from '@/lib/types';
+} from "@mieweb/ui";
+import { Server } from "lucide-react";
+import { api, ApiError } from "@/lib/api";
+import { keys, queries } from "@/lib/queries";
+import { FormPageLayout } from "@/components/FormPageLayout";
+import type { Node } from "@/lib/types";
```
runleveldev: I'm assuming these changes were from an auto-formatter. Please avoid making formatting-only changes since we don't have a consistent formatting standard. I'm open to a PR to enforce formatting standards, but it should be a dedicated PR to avoid expanding feature diffs.

create-a-container/utils/docker-api.js
```diff
@@ -0,0 +1,451 @@
...
+}
+
+function task(kind, id = '') {
+  return `docker:${kind}:${id}`;
+}
+
+class DockerApi {
+  constructor(node = {}) {
+    this.node = node;
+    const dockerHost = node.apiUrl || process.env.DOCKER_HOST || 'unix:///var/run/docker.sock';
```
runleveldev: This should _only_ be node.apiUrl. The fallbacks seem unreachable, since a provisionable node is guarded on the existence of node.apiUrl. If somehow this code is reached wthout the node having an apiUrl, this should throw due to violated preconditions.

create-a-container/utils/docker-api.js
```diff
@@ -0,0 +1,451 @@
+const axios = require('axios');
+
+const MANAGER_NODE_LABEL = 'manager-os.node-id';
+const MANAGER_CONTAINER_LABEL = 'manager-os.container-id';
+
+function parseDockerHost(host) {
+  const raw = host || process.env.DOCKER_HOST || 'unix:///var/run/docker.sock';
```
runleveldev: This fallback is the same as the one below and has the same issue: the fallbacks should be unreachable. I'll allow that the function should validate it's inputs (since this isn't typescript), but it should throw with precondition failed if `typeof host !== 'string'` (or similar).

create-a-container/utils/docker-api.js
```diff
@@ -0,0 +1,451 @@
...
+    return {
+      baseURL: `http://${url.host}`,
+    };
+  }
+
+  if (raw.startsWith('http://') || raw.startsWith('https://')) {
+    return {
+      baseURL: raw.replace(/\/$/, ''),
+    };
+  }
```
runleveldev: Similar to my earlier comments: this chunk of code allows for "bad" URLs and corrects them, such as by stripping path components. Rather than adjusting that here, the API server should validate that the URL is valid before it's ever saved into the database throwing a 4xx error if the user provided an invalid URL. The same validation code could be used here as a safetynet i.e.: `if (!isValid(host)) throw`.

create-a-container/utils/docker-api.js
```diff
@@ -0,0 +1,451 @@
+const axios = require('axios');
+
+const MANAGER_NODE_LABEL = 'manager-os.node-id';
+const MANAGER_CONTAINER_LABEL = 'manager-os.container-id';
```
runleveldev: We've been using the namespace of `org.mieweb.opensource-server`, these should follow that pattern, such as `org.mieweb.opensource-server.node-id`.

create-a-container/utils/docker-api.js
```diff
@@ -0,0 +1,451 @@
...
+      params: options.params,
+      data: options.data,
+      responseType: options.responseType,
+    });
+    return response.data;
+  }
+
+  labels(vmid) {
+    return {
+      [MANAGER_NODE_LABEL]: String(this.node.id || this.node.name || 'docker'),
```
runleveldev: Again, these are completely unnessecary fallbacks. a Node without an id is not a valid state for the process to ever be in.

create-a-container/utils/docker-api.js
```diff
@@ -0,0 +1,451 @@
...
+      existing
+        .map((c) => c.Labels?.[MANAGER_CONTAINER_LABEL])
+        .filter(Boolean)
+        .map(Number),
+    );
+
+    let candidate = Date.now() % 1000000000;
+    while (used.has(candidate)) candidate += 1;
+    return candidate;
+  }
```
runleveldev, line 169: Docker already has a "Container ID" concept based on Sha256 hashes. I would like to use that here so that API requests can refer to the container by it's real id rather than having to look it up by label every time.

create-a-container/utils/docker-api.js
```diff
@@ -0,0 +1,451 @@
...
+
+  async nodes() {
+    const info = await this.request('get', '/info');
+    return [
+      {
+        node: this.node.name || info.Name || 'docker',
+        status: 'online',
+        type: 'node',
+      },
+    ];
```
runleveldev: Node name is required so these fallbacks are unnessecary. This could just be a dummy function and only support 1 node, but I'd like to have this support a docker swarm using the /nodes route.

https://docs.docker.com/reference/api/engine/version/v1.55/#tag/Node/operation/NodeList

create-a-container/utils/docker-api.js
```diff
@@ -0,0 +1,451 @@
...
+  }
+
+  async nodeNetwork() {
```
runleveldev: In a swarm, https://docs.docker.com/reference/api/engine/version/v1.55/#tag/Node/operation/NodeInspect can be used to get some networking info (specifically the IP address which we care about). If the node is not in a swarm this will error so we can fallback to the current behavior.

create-a-container/utils/docker-api.js
```diff
@@ -0,0 +1,451 @@
...
+        total: 0,
+        avail: 0,
+        used: 0,
+      },
+    ];
+  }
+
+  async storageContents() {
+    return [];
+  }
```
runleveldev, line 264: This is used by the `create-container.js` job to check for cached images. It might be sufficient to rely on pullOciImage below to handle caching since docker is better at it. If so, please note that in a comment near this function.

create-a-container/utils/docker-api.js
```diff
@@ -0,0 +1,451 @@
...
+      stream.on('error', reject);
+    });
+
+    return task('pull', image);
+  }
+
+  async createLxc(node, options = {}) {
+    if (!this.lastPulledImage) {
+      throw new Error('No Docker image has been pulled for this create operation');
+    }
```
runleveldev, line 289: This is an unnecessary check. Images might be cached for example.

create-a-container/utils/docker-api.js
```diff
@@ -0,0 +1,451 @@
...
+    const container = await this.findContainerByVmid(vmid);
+    await this.request('delete', `/containers/${container.Id}`, {
+      params: { force: !!force },
+    });
+    return { data: null };
+  }
+
+  async waitForTask() {
+    return { status: 'stopped', exitstatus: 'OK' };
+  }
```
runleveldev, line 480: Is this a no-op because all calls to the Docker API are synchronous? If so, please note that in a comment near this function.

create-a-container/utils/docker-api.js
```diff
@@ -0,0 +1,451 @@
...
+      }
+    }
+
+    return null;
+  }
+
+  async getLxcNetworkInfo(node, vmid) {
+    return {
+      macAddress: await this.getLxcMacAddress(node, vmid),
+      ipv4Address: await this.getLxcIpAddress(node, vmid),
```
runleveldev: Doing this this way creates 2 round-trips to the API. Fine for local, could get painful if remote. You should be able to query the API once for both these pieces of information.

create-a-container/models/node.js
```diff
@@ -24,6 +25,30 @@ module.exports = (sequelize, DataTypes) => {
       });
     }
 
+    hasApiAccess() {
+      if (this.nodeType === 'dummy') return true;
+      if (this.nodeType === 'docker') return !!this.apiUrl;
+      return !!(this.apiUrl && this.tokenId && this.secret);
+    }
+
+    static provisionableWhere(Sequelize) {
```
runleveldev: Can we import Sequelize into this file? Doesn't make sense to me to pass this as an argument.

runleveldev (changes requested):
The UI displays "Creds missing" for Docker nodes

Image

Running our default templates fail since they use SystemD. Similar to how Proxmox checks for "application containers" and make the nessecary adjustments, this needs to check for "system containers" and make the nessacary adjustments (such as setting `privileged: true`)

```
7a542e3b9e12   ghcr.io/mieweb/opensource-server/base:latest            "/sbin/init"             19 seconds ago   Exited (255) 17 seconds ago                hollow-urban-otter
```

create-a-container/utils/docker-api.js
ArshSSandhu: @runleveldev I checked this before changing it. `Containers.containerId` is currently `INTEGER.UNSIGNED` in the model/migrations, so it looks like it is still treated as a numeric Proxmox-style VMID. Since Docker's real container ID is a string/hash, this may require either changing `containerId` to a string or adding a separate provider/external container ID field. Do you prefer one of those approaches, or should I keep the current label lookup until that schema change is handled separately?

create-a-container/client/src/pages/nodes/NodeFormPage.tsx
ArshSSandhu: Fixed in `593a4eb` — `fix(nodes): remove node form formatting noise`.

create-a-container/utils/docker-api.js
ArshSSandhu: Fixed in `05b05eb` — `fix(docker): require node api url`.

create-a-container/utils/docker-api.js
ArshSSandhu: Fixed in `25437b6` — `fix(docker): validate Docker host format`.

create-a-container/utils/docker-api.js
ArshSSandhu: Fixed in `25437b6` and `9afd241` — Docker host validation is shared by `DockerApi` and the node API now rejects invalid Docker hosts before saving.

create-a-container/utils/docker-api.js
ArshSSandhu: Fixed in `b5de6b8` — `fix(docker): use org label namespace`.

create-a-container/utils/docker-api.js
ArshSSandhu: Fixed in `c28a689` — `fix(docker): remove invalid node fallbacks`.

create-a-container/utils/docker-api.js
ArshSSandhu: Fixed in `eb4601a` — `feat(docker): list swarm nodes when available`.

create-a-container/utils/docker-api.js
ArshSSandhu: Fixed in `ff8f675` — `feat(docker): read swarm node network info`.

create-a-container/utils/docker-api.js
ArshSSandhu: Fixed in `5e42932` — `docs(docker): explain image cache handling`.

create-a-container/utils/docker-api.js
ArshSSandhu: Fixed in `b652af3` — `fix(docker): allow cached images during create`.

create-a-container/utils/docker-api.js
ArshSSandhu: Fixed in `4237179` — `docs(docker): explain task wait no-op`.

create-a-container/utils/docker-api.js
ArshSSandhu: Fixed in `dcde503` — `fix(docker): reuse network inspect result`.

create-a-container/models/node.js
ArshSSandhu: Fixed in `c7fd90f` — `fix(nodes): keep provisionable query operators in model`.

ArshSSandhu:
The Docker credential badge was fixed in `15ffc48` — `fix(nodes): handle Docker credential badge`.

The systemd/default template handling was fixed in `e3e9465` — `fix(docker): support system containers`.


I pushed the requested review fixes as separate commits so each item is easier to inspect.

Testing run:
- backend `node --check` commands
- frontend `npm run type-check` from `create-a-container/client`
- `git diff --check`

I replied on the individual threads with the commit that addresses each item. I left the Docker real container ID thread unresolved because `Containers.containerId` is currently `INTEGER.UNSIGNED`, so using Docker’s real string/hash ID needs a schema decision.

create-a-container/utils/docker-api.js
```diff
@@ -205,8 +205,29 @@ class DockerApi {
...
+        ? [
+            {
+              iface: 'docker-swarm',
+              type: 'docker',
+              active: true,
+              address,
+            },
+          ]
+        : [];
+    } catch (err) {
```
runleveldev, line 216: This is swallowing all errors, not just errors indicating that the nodes is not part of a swarm. This could cause a swarm-included node to improperly present as a not swarm node due to communication failures rather than throwing as expected.

create-a-container/utils/docker-api.js
runleveldev: @ArshSSandhu please drop the commit that fixed this. I misunderstood the behavior of the image parsing and given the existing behavior `this.lastPulledImage` should always be set during this operation and docker image caching would already apply properly based on image reference.

create-a-container/utils/docker-api.js
```diff
@@ -88,6 +88,22 @@ function task(kind, id = '') {
...
+    : options.entrypoint || '';
+
+  const image = options.ostemplate || options.reference || '';
+
+  return (
+    entrypoint.includes('/sbin/init') ||
+    entrypoint.includes('systemd') ||
+    image.includes('/base:') ||
+    image.includes('/base@') ||
+    image.endsWith('/base')
```
runleveldev: This is a bad hueristic. Stick with what you've got based on the entrypoint.

create-a-container/routers/api/v1/nodes.js
```diff
@@ -35,6 +37,28 @@ async function loadSite(req) {
...
 }
 
+function normalizeNodeType(nodeType) {
+  return nodeType || 'proxmox';
+}
+
+function validateNodeInput({ nodeType, apiUrl }) {
+  const type = normalizeNodeType(nodeType);
+
+  if (!['proxmox', 'docker', 'dummy'].includes(type)) {
```
runleveldev, line 47: Make sure the OpenAPI spec is updated as well.

runleveldev (changes requested):

create-a-container/utils/docker-api.js
runleveldev: @ArshSSandhu feel free to migrate containerId to a varchar or similar so that this will work, we can always serialize proxmox's numeric IDs as strings.

create-a-container/utils/docker-api.js
ArshSSandhu: Fixed in `4b3f6bd` — `fix(docker): restore pulled image requirement`.

I restored the previous behavior so `createLxc()` requires `this.lastPulledImage` again.

create-a-container/utils/docker-api.js
ArshSSandhu: Fixed in `fdcbea8` — `fix(docker): only ignore non-swarm network errors`.

The fallback now only returns `[]` for the expected non-swarm statuses and rethrows other Docker/API errors.

create-a-container/utils/docker-api.js
ArshSSandhu: Fixed in `ebb9fb2` — `fix(docker): detect system containers by entrypoint`.

I removed the image-name heuristic and kept system-container detection based on the entrypoint.

create-a-container/routers/api/v1/nodes.js
ArshSSandhu: Fixed in `f886191` — `docs(api): document Docker node fields`.

The OpenAPI spec now documents `nodeType`, Docker host values, and the node create/update request schema.

create-a-container/utils/docker-api.js
ArshSSandhu: Fixed in `e65f467` and `fcd4718`.

`containerId` is now string-backed for provider IDs, and Docker operations now address containers by Docker's real container ID instead of looking them up by the manager label.

ArshSSandhu:
@runleveldev 

I pushed the latest review fixes as separate commits:

- `4b3f6bd` — restored the pulled-image requirement
- `fdcbea8` — only ignore expected non-swarm network errors
- `ebb9fb2` — removed the image-name system-container heuristic
- `f886191` — updated OpenAPI for Docker node fields
- `e65f467` — migrated `containerId` to string-backed provider IDs
- `fcd4718` — addressed Docker containers by Docker's real container ID

I also ran the backend syntax checks, frontend type-check, and `git diff --check`.

create-a-container/utils/docker-api.js
```diff
@@ -0,0 +1,536 @@
...
+      Image: inspect.Config.Image,
+      Hostname: inspect.Config.Hostname,
+      Labels: {
+        ...(inspect.Config.Labels || {}),
+        [MANAGER_NODE_LABEL]: String(this.node.id),
+      },
+      Env: envObjectToArray(env),
+      Entrypoint: entrypoint,
+      HostConfig: {
+        ...(inspect.HostConfig || {}),
```
runleveldev, line 400: I'm getting this error when trying to run this against a podman-based Docker-compatible socket.

```
API Error Details: {
  "cause": "NanoCpus conflicts with CpuPeriod and CpuQuota",
  "message": "make cli opts(): NanoCpus conflicts with CpuPeriod and CpuQuota",
  "response": 500
}
```

I'd like to maintain portability between multiple container engines. Should be as simple as filtering `CpuPeriod` and `CpuQuota` out of this object since we're standardizing on NanoCpus.

runleveldev (changes requested):
@ArshSSandhu this was the only thing found in my final QA round. Once it's fixed I think this is ready to merge.

create-a-container/utils/docker-api.js
ArshSSandhu: Fixed in `<commit-sha>` — `fix(docker): remove conflicting CPU options`.

The recreation path now removes `CpuPeriod` and `CpuQuota` from the copied `HostConfig` so Docker-compatible engines such as Podman can use the standardized `NanoCpus` value without conflicting CPU settings.

ArshSSandhu:
@runleveldev I fixed it, let me know if anything else is needed.

runleveldev (dismissed):
@ArshSSandhu I pushed one commit because it was such a small fix by the time I knew what the issue was I already had it fixed. I'm going to approve this, let me know that you've reviewed what I did then I'll merge.

ArshSSandhu (commented):
@runleveldev I think this should compare String(r.vmid) === String(vmid) instead of strict equality, since containerId is now stored as a string but Proxmox snapshots may still return numeric VMIDs.

runleveldev:
> @runleveldev I think this should compare String(r.vmid) === String(vmid) instead of strict equality, since containerId is now stored as a string but Proxmox snapshots may still return numeric VMIDs.

Good catch. My commit fixed the issue for docker containers but just moved it to affecting Proxmox containers.

ArshSSandhu:
@runleveldev I fixed it. This PR should be ready to merge.

runleveldev (approved):

PR #400, opened 2026-07-09
fix(cleanup-preview): use correct secrets names

runleveldev:
# Copilot Summay

This pull request makes a small update to the workflow configuration by changing the secret names used for the `api_key` and `api_url` in the `.github/workflows/build-images.yml` file to use more generic environment variable names.


cmyers-mieweb (approved):
Perfect

PR #407, opened 2026-07-15
Replace pull-config with TypeScript agent

### LINKED ISSUE #357: Agent check-in
Currently, the API server (manager/create-a-container) knows almost nothing about site agents (pull-config clients). Agents are essentially stateless, just querying the API server for configs per the rules of pull-config. What I'd like to do instead is have each agent "check-in" with the manager. This will be on the same schedule as pull-config, but will send additional data, including system info (hostname, ip addresses, etc) and last sync status (success/failure).

Instead of relying on ETags, the API server should respond to the check-in with the services that need updated. This should reduce the overall request load as well and enable caching in a future update.

This will then allow admins to see what agents are active for a site, when they last checked in, their current health. It will also allow the API server to make intelligent decisions about other containers based on the agent status.

cmyers-mieweb:
Resolves: https://github.com/mieweb/opensource-server/issues/357

Replaces the cron-driven bash pull-config system with a TypeScript Node.js agent that runs every 30 seconds via a systemd timer.

Key changes:
- New `agent/` package: TypeScript sources (config, state, api, system, apply), EJS templates moved from create-a-container/views/, systemd service+timer units
- Manager: new `Agents` DB model/migration, `POST /api/v1/agents` check-in endpoint with ETag/304 support, `GET /api/v1/agents` admin list, `utils/agent-config.js` config snapshot builder
- Client: AgentsListPage with service health badges, sidebar nav entry, query/types
- Removes: pull-config bash scripts, EJS template rendering from manager (templates router, ejs dependency, views engine setup)
- Updates docs, Dockerfiles, Makefiles, and OpenAPI spec


agent/src/config.ts
```diff
@@ -0,0 +1,23 @@
...
+  const siteId = parseInt(env.SITE_ID ?? '', 10);
+  const managerUrl = env.MANAGER_URL;
+  if (!Number.isInteger(siteId) || !managerUrl) {
+    throw new Error('SITE_ID and MANAGER_URL must be set in the environment');
+  }
+  return {
+    siteId,
+    managerUrl: managerUrl.replace(/\/+$/, ''),
+    apiKey: env.API_KEY || undefined,
+    stateDir: env.STATE_DIR || '/var/lib/opensource-agent',
```
runleveldev: Use `STATE_DIRECTORY` and set `StateDirectory=opensource-agent` in the service file to let SystemD automatically manage it's lifecycle. This would also remove the need to make the directory in the Makefile install target.

agent/src/state.ts
```diff
@@ -0,0 +1,29 @@
...
+
+function stateFile(stateDir: string): string {
+  return path.join(stateDir, 'state.json');
+}
+
+export function loadState(stateDir: string): AgentState {
+  try {
+    const raw = JSON.parse(fs.readFileSync(stateFile(stateDir), 'utf8'));
+    return { lastApply: {}, ...raw };
+  } catch {
```
runleveldev: I don't like unqualified catch. Only catch known errors (file missing, failed to parse, etc) and throw anything unexpected.

agent/src/state.ts
```diff
@@ -0,0 +1,29 @@
...
+    return { lastApply: {} };
+  }
+}
+
+export function saveState(stateDir: string, state: AgentState): void {
```
runleveldev: Just a thought: why don't we make a State class that has load and save methods rather than handling this procedurally?

agent/src/system.ts
```diff
@@ -0,0 +1,26 @@
...
+      if (iface.family === 'IPv4' && !iface.internal) return iface.address;
+    }
+  }
+  return null;
+}
+
+/** systemd ActiveState for a unit (active, inactive, failed, ...). */
+export function getServiceState(unit: string): string {
+  try {
+    const out = execFileSync('systemctl', ['show', '--property=ActiveState', '--value', unit], {
```
runleveldev: Do not exec subprocesses: SystemD supposts a Unix-socket based API (D-Bus) which should be used for these kinds of operations.

agent/src/apply.ts
```diff
@@ -0,0 +1,129 @@
...
+    unit: 'nginx',
+    async render(config) {
+      return [{
+        dest: '/etc/nginx/nginx.conf',
+        content: await renderTemplate('nginx.conf.ejs', config.nginx),
+      }];
+    },
+    test: ['nginx', '-t'],
+    reload() {
+      run(['systemctl', 'reload-or-restart', 'nginx']);
```
runleveldev: I won't complain about the `test` command, but the reload shoudl use the SystemD DBus API rather than subcommands.

agent/src/apply.ts
```diff
@@ -0,0 +1,129 @@
...
+      rollback();
+      console.error(`${svc.unit}: config test failed, rolled back: ${(err as Error).message}`);
+      return 'failure';
+    }
+  }
+
+  try {
+    svc.reload(changed);
+  } catch (err) {
+    console.error(`${svc.unit}: reload failed: ${(err as Error).message}`);
```
runleveldev: Should a failed reload also rollback?

create-a-container/routers/api/v1/agents.js
```diff
@@ -0,0 +1,80 @@
...
+// The manager's own agent checks in over localhost without credentials
+// (bootstrap: no site, no API key exist yet). Remote agents authenticate
+// with an admin API key.
+function agentAuth(req, res, next) {
+  if (isLocalhostRequest(req)) return next();
+  return apiAuth(req, res, (err) => {
+    if (err) return next(err);
+    return apiAdmin(req, res, next);
+  });
+}
```
runleveldev: Was the `requireLocalhostOrAdmin` middleware not sufficient?

create-a-container/routers/api/v1/agents.js
```diff
@@ -0,0 +1,80 @@
...
+      lastCheckinAt: new Date(),
+    });
+  }
+
+  const config = await buildAgentConfig(parsedSiteId);
+  const etag = computeConfigEtag(config);
+  res.set('ETag', etag);
+  if (req.get('If-None-Match') === etag) {
+    return res.status(304).end();
+  }
```
runleveldev, line 59: Fact check me on thin, but I'm pretty sure Express already handles ETag logic. I was using it for the original pull-config setup but that might only work with rendering templates.

create-a-container/routers/api/v1/agents.js
```diff
@@ -0,0 +1,80 @@
...
+  const now = Date.now();
+  return ok(res, agents.map((a) => ({
+    id: a.id,
+    siteId: a.siteId,
+    siteName: a.site?.name || null,
+    hostname: a.hostname,
+    ipv4Address: a.ipv4Address,
+    services: a.services,
+    lastCheckinAt: a.lastCheckinAt,
+    online: !!a.lastCheckinAt && now - new Date(a.lastCheckinAt).getTime() <= ONLINE_WINDOW_MS,
```
runleveldev: I do want the time-since-checkin to be computed serverside to avoid client-side clock drift in the UI, but I think this field should just be that delta rather than a boolean is-online and the UI (or even just the human admin) could decide whether there's an issue based on that.

mie-opensource-landing/docs/admins/deploying-agents.md
```diff
@@ -69,33 +69,34 @@ lxc.environment = API_KEY=<admin-api-key>
...
 
-# Test manually
-/etc/pull-config.d/nginx
-/etc/pull-config.d/dnsmasq-conf
-/etc/pull-config.d/dnsmasq-dhcp-hosts
-/etc/pull-config.d/dnsmasq-hosts
-/etc/pull-config.d/dnsmasq-dhcp-opts
-/etc/pull-config.d/dnsmasq-servers
+# Run a check-in manually
+node /opt/opensource-server/agent/dist/index.js
```
runleveldev: `systemctl start opensource-agent` to ensure the systemd environment is respected.

mie-opensource-landing/docs/developers/agent.md
```diff
@@ -0,0 +1,99 @@
...
+│   ├── api.ts               # Check-in HTTP client
+│   └── apply.ts             # Managed services: render/test/apply/reload
+├── templates/               # EJS templates rendered locally
+│   ├── nginx.conf.ejs
+│   └── dnsmasq/             # conf, dhcp-hosts, hosts, dhcp-opts, servers
+├── contrib/
+│   ├── systemd/             # opensource-agent.service + .timer (30s)
+│   └── postinstall.sh       # Enables the timer on package install
+└── Makefile                 # Builds the opensource-agent package (see Release Pipeline)
+```
```
runleveldev: Don't do this, sucks to keep up-to-date on changes.

runleveldev (changes requested):

agent/src/config.ts
cmyers-mieweb: Done in [`15f0e4a`](https://github.com/mieweb/opensource-server/commit/15f0e4aecfb308f6398f9b505a16ef400aa2926e) — the unit now sets `StateDirectory=opensource-agent`, `config.ts` reads `$STATE_DIRECTORY`, and the Makefile no longer creates the directory.

agent/src/state.ts
cmyers-mieweb: Fixed in [`15f0e4a`](https://github.com/mieweb/opensource-server/commit/15f0e4aecfb308f6398f9b505a16ef400aa2926e) — load now only handles `ENOENT` (first run) and `SyntaxError` (corrupt file, logged and ignored); anything unexpected is rethrown.

agent/src/apply.ts
```diff
@@ -0,0 +1,129 @@
...
+    reload() {
+      run(['systemctl', 'reload-or-restart', 'nginx']);
+    },
+  },
+  {
+    unit: 'dnsmasq',
+    async render(config) {
+      if (!config.site) return null;
+      const data = { site: config.site };
+      return [
```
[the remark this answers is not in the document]
cmyers-mieweb: Fixed in [`15f0e4a`](https://github.com/mieweb/opensource-server/commit/15f0e4aecfb308f6398f9b505a16ef400aa2926e) — the dnsmasq render now skips (returns `null`) until `internalDomain`, `dhcpRange`, `subnetMask`, `gateway` and `dnsForwarders` are all set.

agent/src/types.ts
```diff
@@ -0,0 +1,74 @@
...
+}
+
+export interface SiteInfo {
+  id: number;
+  name: string;
+  internalDomain: string;
+  dhcpRange: string;
+  subnetMask: string;
+  gateway: string;
+  dnsForwarders: string;
```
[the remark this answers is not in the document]
cmyers-mieweb: Fixed in [`15f0e4a`](https://github.com/mieweb/opensource-server/commit/15f0e4aecfb308f6398f9b505a16ef400aa2926e) — `SiteInfo` fields are now `string | null` to match the manager's Site model, and the dnsmasq renderer guards on them.

agent/src/index.ts
```diff
@@ -0,0 +1,65 @@
...
+      state.lastApply[svc.unit] = await applyService(svc, result.config);
+    }
+
+    // The ETag is saved even after a failed apply: a rejected config won't
+    // fix itself without a server-side change (which changes the ETag), and
+    // the failure has been reported via lastApply.
+    state.etag = result.etag;
+    saveState(cfg.stateDir, state);
+  }
+}
```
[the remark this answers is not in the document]
cmyers-mieweb: Fixed in [`15f0e4a`](https://github.com/mieweb/opensource-server/commit/15f0e4aecfb308f6398f9b505a16ef400aa2926e) — when `MAX_PASSES` is exhausted the agent now checks in once more so the final pass' apply results reach the manager instead of going stale until the next timer run.

agent/src/state.ts
cmyers-mieweb: Done in [`15f0e4a`](https://github.com/mieweb/opensource-server/commit/15f0e4aecfb308f6398f9b505a16ef400aa2926e) — `state.ts` is now a `State` class with a static `load()` and an instance `save()`.

mie-opensource-landing/docs/developers/agent.md
cmyers-mieweb: Removed in [`15f0e4a`](https://github.com/mieweb/opensource-server/commit/15f0e4aecfb308f6398f9b505a16ef400aa2926e), replaced with a one-sentence pointer to the relevant directories.

create-a-container/routers/api/v1/agents.js
cmyers-mieweb: It was — I'd missed that `requireAuth` also accepts Bearer API keys. Switched to `requireLocalhostOrAdmin` and dropped the custom `agentAuth` in [`15f0e4a`](https://github.com/mieweb/opensource-server/commit/15f0e4aecfb308f6398f9b505a16ef400aa2926e).

agent/src/apply.ts
cmyers-mieweb: Good catch — yes. In [`15f0e4a`](https://github.com/mieweb/opensource-server/commit/15f0e4aecfb308f6398f9b505a16ef400aa2926e) a failed reload now rolls back the files and best-effort re-reloads, so the service keeps running the last known-good config; the failure is still reported at the next check-in.

mie-opensource-landing/docs/admins/deploying-agents.md
cmyers-mieweb: Done in [`15f0e4a`](https://github.com/mieweb/opensource-server/commit/15f0e4aecfb308f6398f9b505a16ef400aa2926e) — the docs now use `systemctl start opensource-agent.service`.

agent/src/system.ts
cmyers-mieweb: Done in [`15f0e4a`](https://github.com/mieweb/opensource-server/commit/15f0e4aecfb308f6398f9b505a16ef400aa2926e) — `system.ts` now talks to `org.freedesktop.systemd1` on the system bus (`LoadUnit` + `ActiveState` property) via `@jellybrick/dbus-next`. I used that maintained fork because upstream `dbus-next` pulls in a vulnerable `usocket`/`node-gyp`/`request` chain (3 criticals in `npm audit`); the fork has zero findings. Note this makes the D-Bus system bus a runtime requirement in the containers — the Proxmox Debian template ships dbus, but worth a smoke test.

create-a-container/routers/api/v1/agents.js
cmyers-mieweb: Done in [`15f0e4a`](https://github.com/mieweb/opensource-server/commit/15f0e4aecfb308f6398f9b505a16ef400aa2926e) — the endpoint now returns a server-computed `secondsSinceCheckin` (`null` when never checked in) instead of the boolean; the client decides the offline threshold and renders the relative time from the delta, so nothing depends on the browser clock.

agent/src/apply.ts
cmyers-mieweb: Done in [`15f0e4a`](https://github.com/mieweb/opensource-server/commit/15f0e4aecfb308f6398f9b505a16ef400aa2926e) — reloads now go through the D-Bus API: `ReloadOrRestartUnit`/`RestartUnit` (waiting on the `JobRemoved` signal for the job result) and `KillUnit` for the SIGHUP. The `test` commands remain execs as discussed.

create-a-container/routers/api/v1/agents.js
cmyers-mieweb: Fact-checked: Express's automatic ETag/304 handling (`res.send` + `req.fresh`) only applies to GET/HEAD — `req.fresh` is always false for POST, which is what the check-in uses. That's why it worked for the old GET-based pull-config templates but not here. Added a comment explaining this in [`15f0e4a`](https://github.com/mieweb/opensource-server/commit/15f0e4aecfb308f6398f9b505a16ef400aa2926e).

runleveldev:
So far unable to root cause. Bootstrapping in docker compose is failing because the agent fails to apply the initial config.

```
Jul 21 19:20:01 manager node[914]: nginx: reload failed, rolled back: fast_xml_builder.default is not a constructor
```

To reproduce, set `MANAGER_TAG=sha-ca620d0` in .env, `make -C agent build`, then `docker compose up -d`

agent/src/system.ts
cmyers-mieweb: Follow-up in [`dd77c50`](https://github.com/mieweb/opensource-server/commit/dd77c50e6a1b0436c55a5ccde116d91b6b34fe02): the jellybrick fork turned out to have a broken ESM-interop shim in its CommonJS bundle — its `MessageBus` constructor calls `new fast_xml_builder.default(...)` where `__toESM()` has set `.default` to the whole module namespace, so every bus operation failed with `fast_xml_builder.default is not a constructor` during compose bootstrap. Swapped to `@particle/dbus-next` (the original dbus-next codebase, native CJS, with the vulnerable `usocket`/`node-gyp`/`request` chain removed) — same API, `npm audit` still clean. Verified end-to-end in docker compose: nginx + dnsmasq apply, D-Bus reload/restart/SIGHUP all work, both services active with `lastApply: success`.

cmyers-mieweb: Correct commit link (typo in the SHA above): [`dd77c50`](https://github.com/mieweb/opensource-server/commit/dd77c50c87c5d106c7e50b17e3b031e20554ae32)

cmyers-mieweb:
> So far unable to root cause. Bootstrapping in docker compose is failing because the agent fails to apply the initial config.
> 
> ```
> Jul 21 19:20:01 manager node[914]: nginx: reload failed, rolled back: fast_xml_builder.default is not a constructor
> ```
> 
> To reproduce, set `MANAGER_TAG=sha-ca620d0` in .env, `make -C agent build`, then `docker compose up -d`

I pushed a new commit https://github.com/mieweb/opensource-server/pull/407/commits/dd77c50c87c5d106c7e50b17e3b031e20554ae32 that fixes the issue. The root cause copilot found was related to the following:
its `MessageBus` constructor calls new `fast_xml_builder.default(...)` where` __toESM()` has set `.default` to the whole module namespace, so every bus operation failed with `fast_xml_builder.default` is not a constructor during compose bootstrap. Swapped to `@particle/dbus-next` (the original dbus-next codebase, native CJS, with the vulnerable usocket/node-gyp/request chain removed); same API, npm audit still clean. Verified end-to-end in docker compose: nginx + dnsmasq apply, D-Bus reload/restart/SIGHUP all work, both services active with lastApply: success.

image

image

runleveldev (dismissed):
I've ran several QA tests locally in the compose setup, including bootstrapping, killing services, and updating an existing system via the DEB packages. Everything seems in order. @cmyers-mieweb can you address Copilot's most recent reviews then we can merge.

create-a-container/routers/api/v1/agents.js
```diff
@@ -0,0 +1,76 @@
...
+const { requireLocalhostOrAdmin } = require('../../../middlewares');
+const { apiAuth, apiAdmin, asyncHandler, ok, fail } = require('../../../middlewares/api');
+const { buildAgentConfig, computeConfigEtag } = require('../../../utils/agent-config');
+
+const router = express.Router();
+
+// Check-in auth: the manager's own agent checks in over localhost without
+// credentials (bootstrap: no site, no API key exist yet); remote agents
+// authenticate with an admin API key.
+router.post('/', requireLocalhostOrAdmin, asyncHandler(async (req, res) => {
```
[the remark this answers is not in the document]
cmyers-mieweb: resolved in https://github.com/mieweb/opensource-server/pull/407/commits/cfcfda557c9b2c1d5f3070b755658627295b2ce5

agent/src/api.ts
```diff
@@ -0,0 +1,34 @@
...
+): Promise<CheckinResult> {
+  const headers: Record<string, string> = { 'Content-Type': 'application/json' };
+  if (etag) headers['If-None-Match'] = etag;
+  if (cfg.apiKey) headers['Authorization'] = `Bearer ${cfg.apiKey}`;
+
+  const res = await fetch(`${cfg.managerUrl}/api/v1/agents`, {
+    method: 'POST',
+    headers,
+    body: JSON.stringify(body),
+  });
```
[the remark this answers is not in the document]
cmyers-mieweb: resolved in https://github.com/mieweb/opensource-server/pull/407/commits/cfcfda557c9b2c1d5f3070b755658627295b2ce5

create-a-container/middlewares/index.js
```diff
@@ -90,8 +89,14 @@ function requireLocalhostOrAdmin(req, res, next) {
 
   const realIp = req.get('X-Real-IP');
 
-  // If direct connection is from localhost and no non-localhost X-Real-IP, allow through
-  if (isLocalhost(directIp) && (!realIp || isLocalhost(realIp))) {
+  return isLocalhost(directIp) && (!realIp || isLocalhost(realIp));
```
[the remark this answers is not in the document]
cmyers-mieweb: resolved in https://github.com/mieweb/opensource-server/pull/407/commits/cfcfda557c9b2c1d5f3070b755658627295b2ce5

create-a-container/client/src/pages/agents/AgentsListPage.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+function OnlineBadge({ agent }: { agent: Agent }) {
+  const online =
+    agent.secondsSinceCheckin !== null && agent.secondsSinceCheckin <= OFFLINE_AFTER_SECONDS;
+  return online ? (
+    <Badge variant="success">Online</Badge>
+  ) : (
+    <Badge variant="danger">Offline</Badge>
+  );
+}
```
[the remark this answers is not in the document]
cmyers-mieweb: resolved in https://github.com/mieweb/opensource-server/pull/407/commits/cfcfda557c9b2c1d5f3070b755658627295b2ce5

cmyers-mieweb:
Remaining copilot suggestions have been implemented in https://github.com/mieweb/opensource-server/pull/407/commits/cfcfda557c9b2c1d5f3070b755658627295b2ce5

runleveldev (approved):
Copilot reviews have been over aggressive. I'm merging this despite them open.

PR #411, opened 2026-07-15
docs(openapi): update openapi schema with missing information

runleveldev:
Used AI to make the OpenAPI schema actually match what the API server uses.

### DIFF
--- a/create-a-container/openapi.v1.yaml
+++ b/create-a-container/openapi.v1.yaml
@@ -23,7 +23,19 @@ info:
 
     Success: `{ "data": <payload>, "meta"?: {...} }`.
     Error:   `{ "error": { "code": "...", "message": "...", "fields"?: {...} } }`.
-  version: 1.0.0
+
+    ## Common errors (any endpoint)
+
+    Not repeated on every operation below:
+
+    - `401 unauthorized` — missing/invalid session or API key (any authenticated endpoint).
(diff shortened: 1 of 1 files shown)


create-a-container/openapi.v1.yaml
```diff
@@ -153,31 +262,163 @@ components:
...
+          description: Proxmox API token secret. Never returned (see `hasSecret`). On update, blank/omitted keeps the existing secret.
         tlsVerify: { type: boolean, nullable: true }
-        imageStorage: { type: string }
-        volumeStorage: { type: string }
-        networkBridge: { type: string }
-        nvidiaAvailable: { type: boolean }
+        imageStorage: { type: string, default: local }
+        volumeStorage: { type: string, default: local-lvm }
+        networkBridge: { type: string, default: vmbr0 }
+        nvidiaAvailable: { type: boolean, default: false }
```
cmyers-mieweb, line 275: How do we feel about storing a var to pull available VRAM? If GPU use is under heavy constraint, it might be worthwhile to include it to give transparency of the current available resources.

runleveldev: Out of scope for this PR. The intention here is just to make sure the schema matches reality.

cmyers-mieweb (approved):
LGTM

Issue #421, opened 2026-07-22
Support user-defined shared volumes for containers (replace hardcoded quick_and_dirty mp0)


runleveldev:
## Background

Container creation (`create-a-container/bin/create-container.js`) now attaches a hardcoded shared volume to every created container: the node's template storage is assumed to be mounted at `/mnt/pve/<storage>` (i.e. templates are downloaded to `/mnt/pve/<storage>/template/cache`), and `/mnt/pve/<storage>/volumes/quick_and_dirty` is bind-mounted read-only at `/mnt/quick_and_dirty` via `mp0`.

This is a quick-and-dirty stopgap (see the `TODO` on `buildSharedVolumeMp0()` in `create-a-container/bin/create-container.js`). It should be replaced with a proper feature where users can define their own volumes and attach them to containers with per-volume permissions.

## Proposed scope

- Volume management (CRUD) for named volumes backed by directories on node storage, persisted in the database (scoping TBD: per user, per site, and/or per node)
- Attach one or more volumes at container creation and via reconfigure (`mp0`..`mpN`), including detach
- Per-attachment permissions: read-only vs read-write (and possibly other mount options such as `backup`, quotas/size limits)
- UI/API support in create-a-container (client form + `/api/v1` endpoints + OpenAPI spec)
- Derive the host path from the node's actual storage configuration (query the storage's `path` via the Proxmox API) instead of assuming the `/mnt/pve/<storage>` layout
- Validate volume names (safe path characters, no traversal) and ensure the source directory exists before mounting
- Handle unprivileged container UID/GID mapping so read-write volumes are actually writable from inside containers
- Decide behavior for Docker nodes (bind-mount options are currently ignored by the Docker backend)
- Document privilege requirements: Proxmox restricts raw bind mounts (`mp0` with a host path) to root@pam, so the API credentials used by the manager must allow it

## Current stopgap behavior (to be removed)

- Volume name hardcoded to `quick_and_dirty`
- Always mounted read-only at `/mnt/quick_and_dirty`
- Applied to all containers created on Proxmox nodes (both the OCI image path and the template clone path)

runleveldev:
[...]

PR #422, opened 2026-07-22
feat(containers): mount shared read-only volume into created containers

runleveldev:
## Summary

Adds an `mp0` mount point to every created container: `/mnt/pve/<templateStorage>/volumes/quick_and_dirty` is bind-mounted read-only at `/mnt/quick_and_dirty`.

The host path is derived from the node's resolved template storage, assuming the storage is mounted at `/mnt/pve/<storage>` (i.e. container templates are downloaded to `/mnt/pve/<storage>/template/cache`), so shared volumes live at `/mnt/pve/<storage>/volumes/<name>`.

This is an intentional stopgap: the volume name is hardcoded to `quick_and_dirty`. #421 tracks turning this into a proper feature where users can define their own volumes with per-volume permissions; a `TODO(#421)` is left on the new `buildSharedVolumeMp0()` helper.

## Changes

- New `buildSharedVolumeMp0(templateStorage)` helper in `create-a-container/bin/create-container.js` building the mp0 value (`/mnt/pve/<storage>/volumes/quick_and_dirty,mp=/mnt/quick_and_dirty,ro=1`)
- **OCI image path**: `mp0` added to the `createLxc` options, reusing the already-resolved template storage
- **Template clone path**: template storage is now resolved in this branch too (non-fatal — if no `vztmpl`-capable storage exists, the mount is skipped with a warning instead of failing creation, since cloning doesn't otherwise need one) and `mp0` is applied via the post-clone `updateLxcConfig` call

## Notes

- Proxmox restricts raw bind mounts (`mp0` with a host path) to `root@pam`, so the manager's API credentials must allow this
- Docker nodes silently ignore the option (the Docker backend's `createLxc` only maps known fields), so Docker containers won't get the mount — noted in #421

## Testing

- `node --check create-a-container/bin/create-container.js` passes (no backend test suite exists for these job scripts)

Refs #421

### DIFF
--- a/create-a-container/bin/create-container.js
+++ b/create-a-container/bin/create-container.js
@@ -92,6 +92,28 @@ async function resolveStorage(client, nodeName, preferred, contentType) {
 }
 
 /**
+ * Build the mp0 mount point value for the shared read-only volume.
+ * Assumes the node's template storage is a directory storage mounted at
+ * /mnt/pve/<storage> (i.e. container templates are downloaded to
+ * /mnt/pve/<storage>/template/cache), so shared volumes live at
+ * /mnt/pve/<storage>/volumes/<name>.
+ *
+ * TODO(#421): Replace this hardcoded 'quick_and_dirty' volume with a proper
+ * feature where users can define their own volumes and attach them with
+ * per-volume permissions (read-only/read-write). The host path should be
+ * derived from the storage's actual configured path instead of assuming the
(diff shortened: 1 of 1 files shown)


cmyers-mieweb (approved):
Approved, we will need to loop back on this in the future and make an issue to make it an optional feature if the hardware is not under a ceph infrastructure.

PR #464, opened 2026-08-25
Enhance container management and UI with version info and fixes

### LINKED ISSUE #358: Add version footer + bug report link back
These are missing in the react spa.

### LINKED ISSUE #449: Saving config should *NEVER* restart a container until the user confirms
## Steps

Image

Image

## Plan
* Never restart without a warning
* Most changes should not require a restart

### LINKED ISSUE #453: [Bug]: Jobs Back button is broken
### Current URL

/jobs/<n>

### Username

any

### Version

v2026.8.1

## also back button is broken

Image

Image

### Expected behavior

goes back to /containers

### Actual behavior

goes to unhandled /jobs

### Additional context

Originally in #449

cmyers-mieweb:
Resolves:
https://github.com/mieweb/opensource-server/issues/358
https://github.com/mieweb/opensource-server/issues/449
https://github.com/mieweb/opensource-server/issues/453

This pull request introduces two main improvements: (1) a new application footer that displays the running version and provides a "Report a bug" link pre-filled with contextual information, and (2) a significant change to container update behavior, ensuring containers are only restarted when the user explicitly requests it, never implicitly. It also improves user feedback and API documentation to reflect this behavior, and enhances navigation in the job detail page.

**User Interface Enhancements:**

* Added a new `AppFooter` component to the UI, which displays the running version (linked to its GitHub commit) and a "Report a bug" link that pre-fills the GitHub issue template with the current URL, username, and version. The footer is now included in the main layout. (`AppFooter.tsx`, `AppLayout.tsx`, `auth.ts`, `index.js`) [[1]](diffhunk://#diff-6147fff69cb56edddd971903566c3d9dc38236ff9ff17aa29113998ae916d14eR1-R48) [[2]](diffhunk://#diff-7b19c269f78c3bbddf844f20efe8723cc94db240143562a4a91626ab497c8a58R6) [[3]](diffhunk://#diff-7b19c269f78c3bbddf844f20efe8723cc94db240143562a4a91626ab497c8a58R20) [[4]](diffhunk://#diff-d7148abd0dca74a08c7195af7927048fcbaab7e3be48a435565188d8bc8e21f8R9-R18) [[5]](diffhunk://#diff-d7148abd0dca74a08c7195af7927048fcbaab7e3be48a435565188d8bc8e21f8R29-R30) [[6]](diffhunk://#diff-5436c5c717e5f972e05844facb2a423c1b4990183f5b672b3f65e55092fc96e7R59-R60)
* The job detail page "Back" button now returns to the current site's containers list (or sites list as a fallback), improving navigation. (`JobDetailPage.tsx`) [[1]](diffhunk://#diff-4c780e38255483535be26afcb536442680a7fa64fbb8524eea119e55db5653a9R14) [[2]](diffhunk://#diff-4c780e38255483535be26afcb536442680a7fa64fbb8524eea119e55db5653a9R41-R44) [[3]](diffhunk://#diff-4c780e38255483535be26afcb536442680a7fa64fbb8524eea119e55db5653a9L116-R123)

**Container Update and Restart Behavior:**

* Changed the backend and API so that updating container configuration (environment variables or entrypoint) never implicitly restarts the container. A restart is only performed when explicitly requested by the user. (`containers.js`, `openapi.v1.yaml`) [[1]](diffhunk://#diff-44647bcd3b853c60c4948db454e8deec1dd1e5aa05f1dbdf4e42ef9daa2e0f22L645-R648) [[2]](diffhunk://#diff-bdedb7b5057f4a7c805abbe4ef41948dd186af96f25737d1c806c2c97ba31151L840-R840) [[3]](diffhunk://#diff-bdedb7b5057f4a7c805abbe4ef41948dd186af96f25737d1c806c2c97ba31151L863-R863)
* The API response now includes a `pendingRestart` flag when changes are saved but not yet applied, and the user interface displays clear feedback when changes require a restart to take effect. (`containers.js`, `ContainerFormPage.tsx`, `openapi.v1.yaml`) [[1]](diffhunk://#diff-44647bcd3b853c60c4948db454e8deec1dd1e5aa05f1dbdf4e42ef9daa2e0f22R776-R786) [[2]](diffhunk://#diff-94ee4a3b78795dc7d8b17b4066f261aabd1e48c80b251c9bb8e95f9e9989846aR339) [[3]](diffhunk://#diff-94ee4a3b78795dc7d8b17b4066f261aabd1e48c80b251c9bb8e95f9e9989846aL340-R358) [[4]](diffhunk://#diff-94ee4a3b78795dc7d8b17b4066f261aabd1e48c80b251c9bb8e95f9e9989846aL555-R584) [[5]](diffhunk://#diff-bdedb7b5057f4a7c805abbe4ef41948dd186af96f25737d1c806c2c97ba31151R878)
* When editing a container, if the user chooses to "restart after saving", a confirmation modal is now shown to prevent accidental disruptive restarts. (`ContainerFormPage.tsx`) [[1]](diffhunk://#diff-94ee4a3b78795dc7d8b17b4066f261aabd1e48c80b251c9bb8e95f9e9989846aR235-R238) [[2]](diffhunk://#diff-94ee4a3b78795dc7d8b17b4066f261aabd1e48c80b251c9bb8e95f9e9989846aR383-R393) [[3]](diffhunk://#diff-94ee4a3b78795dc7d8b17b4066f261aabd1e48c80b251c9bb8e95f9e9989846aL384-R413) [[4]](diffhunk://#diff-94ee4a3b78795dc7d8b17b4066f261aabd1e48c80b251c9bb8e95f9e9989846aR855-R889)

**API and Documentation Updates:**

* Updated the OpenAPI documentation to clarify that restarts only occur when explicitly requested, and documented the new `pendingRestart` response property. (`openapi.v1.yaml`) [[1]](diffhunk://#diff-bdedb7b5057f4a7c805abbe4ef41948dd186af96f25737d1c806c2c97ba31151L840-R840) [[2]](diffhunk://#diff-bdedb7b5057f4a7c805abbe4ef41948dd186af96f25737d1c806c2c97ba31151L863-R863) [[3]](diffhunk://#diff-bdedb7b5057f4a7c805abbe4ef41948dd186af96f25737d1c806c2c97ba31151R878)
* The server `/health` endpoint now includes version information for the frontend footer. (`index.js`) [[1]](diffhunk://#diff-5436c5c717e5f972e05844facb2a423c1b4990183f5b672b3f65e55092fc96e7L46-R46) [[2]](diffhunk://#diff-5436c5c717e5f972e05844facb2a423c1b4990183f5b672b3f65e55092fc96e7R59-R60)

### DIFF
--- a/create-a-container/routers/api/v1/containers.js
+++ b/create-a-container/routers/api/v1/containers.js
@@ -642,7 +642,10 @@ router.put(
     const ownerChanged = newOwnerUsername !== null && newOwnerUsername !== container.username;
     const envChanged = !isRestartOnly && container.environmentVars !== envVarsJson;
     const entrypointChanged = !isRestartOnly && container.entrypoint !== newEntrypoint;
-    const needsRestart = forceRestart || envChanged || entrypointChanged;
+    // Never restart implicitly (issue #449): a restart job is enqueued only
+    // when the caller explicitly asks for one. Saved env/entrypoint changes
+    // are applied by reconfigure-container.js on the next restart.
+    const needsRestart = forceRestart;
 
     let restartJob = null;
     const dnsWarnings = [];
@@ -770,11 +773,17 @@ router.put(
       }
     }
 
+    const pendingRestart = !restartJob && (envChanged || entrypointChanged);
     return ok(res, {
       containerId: container.id,
       jobId: restartJob ? restartJob.id : null,
       dnsWarnings,
-      message: restartJob ? 'Container is restarting' : 'Container updated',
+      pendingRestart,
+      message: restartJob
+        ? 'Container is restarting'
+        : pendingRestart
+          ? 'Container updated — changes take effect on the next restart'
+          : 'Container updated',
     });
   }),
 );
--- a/create-a-container/routers/api/v1/index.js
+++ b/create-a-container/routers/api/v1/index.js
@@ -43,7 +43,7 @@ const { isOidcEnabled } = require('../../../utils/oidc');
 const { Setting } = require('../../../models');
 router.get(
   '/health',
-  asyncHandler(async (_req, res) => {
+  asyncHandler(async (req, res) => {
     // The banner is cosmetic — never let a DB hiccup fail the health check.
     let banner = null;
     try {
@@ -56,6 +56,8 @@ router.get(
       isDev: process.env.NODE_ENV !== 'production',
       oidcEnabled: isOidcEnabled(),
       banner,
+      // Cached at startup in app.locals (see app.js); the SPA footer shows it.
+      version: req.app.locals.versionInfo ?? null,
     });
   }),
 );
--- a/create-a-container/routers/api/v1/index.js
+++ b/create-a-container/routers/api/v1/index.js
@@ -43,7 +43,7 @@ const { isOidcEnabled } = require('../../../utils/oidc');
 const { Setting } = require('../../../models');
 router.get(
   '/health',
-  asyncHandler(async (req, res) => {
+  asyncHandler(async (_req, res) => {
     // The banner is cosmetic — never let a DB hiccup fail the health check.
     let banner = null;
     try {
@@ -56,8 +56,6 @@ router.get(
       isDev: process.env.NODE_ENV !== 'production',
       oidcEnabled: isOidcEnabled(),
       banner,
-      // Cached at startup in app.locals (see app.js); the SPA footer shows it.
-      version: req.app.locals.versionInfo ?? null,
     });
   }),
 );
--- a/create-a-container/Makefile
+++ b/create-a-container/Makefile
@@ -41,6 +41,9 @@ deps:
 	npm --prefix client ci
 
 build: deps
+	# Bake the release version into the client bundle (vite define, see
+	# client/vite.config.ts) — the deployed app has no git checkout to ask.
+	npm --prefix client pkg set version="$$(../package-version node)"
 	npm --prefix client run build
 
 # Run the Manager locally against SQLite, with a dummy (mock) hypervisor so
--- a/create-a-container/app.js
+++ b/create-a-container/app.js
@@ -105,11 +105,6 @@ function buildApp({
   const { csrfGuard, jsonErrorHandler } = require('./middlewares/api');
   app.use(csrfGuard);
 
-  // Set version info once at startup in app.locals
-  // Note: Version info is cached at startup. Server restart required to update version.
-  const { getVersionInfo } = require('./utils');
-  app.locals.versionInfo = getVersionInfo();
-
   // --- Mount Routers ---
   const apiV1Router = require('./routers/api/v1');
 
--- a/create-a-container/client/src/app/AppFooter.tsx
+++ b/create-a-container/client/src/app/AppFooter.tsx
@@ -1,0 +1,48 @@
+import { useLocation } from 'react-router';
+import { Bug } from 'lucide-react';
+import { useServerInfo, useSession } from '@/lib/auth';
+
+const REPO_URL = 'https://github.com/mieweb/opensource-server';
+
+/**
+ * App-wide footer showing the running version (linked to its commit) and a
+ * "Report a bug" link that pre-fills the GitHub bug-report template with the
+ * current URL, username, and version.
+ */
+export function AppFooter() {
+  const { data: serverInfo } = useServerInfo();
+  const { data: session } = useSession();
+  const location = useLocation();
+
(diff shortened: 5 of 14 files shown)


create-a-container/routers/api/v1/index.js
```diff
@@ -56,6 +56,8 @@ router.get(
       isDev: process.env.NODE_ENV !== 'production',
       oidcEnabled: isOidcEnabled(),
       banner,
+      // Cached at startup in app.locals (see app.js); the SPA footer shows it.
```
runleveldev: The server-side getVersionInfo relies on the local filesystem the server is running from being a git repository. This assumption does not hold once the software is packaged and deployed to the live system. As it stands, the source of truth for the current version would be the version of the installed deb package, but trying to read that would be overkill. Since the server and client are distributed in the same package, we should bake the version into the client build in the packaging workflow.

1. A release is generated with the version in the tag
2. `.github/workflows/release.yml` builds `images/builder` which generates the deb packages
3. Within `images/builder`, `make deb` is ran from the repo root
4. `make deb` delegates to the Makefile in each major component, the create-a-container Makefile runs `npm build` for the client component
5. `npm build` in the client delegates to `vite build`

https://github.com/mieweb/ozwell-studio also uses vite build and a Github-releases driven versioning. This sets the version in the package.json just before the final build and bakes the value into the final artifact so there is no live-query required.

Implement this same idea in the build workflow here, then remove the existing logic to interogate the version from the local git repo.

create-a-container/routers/api/v1/containers.js
```diff
@@ -642,7 +642,10 @@ router.put(
     const ownerChanged = newOwnerUsername !== null && newOwnerUsername !== container.username;
     const envChanged = !isRestartOnly && container.environmentVars !== envVarsJson;
     const entrypointChanged = !isRestartOnly && container.entrypoint !== newEntrypoint;
-    const needsRestart = forceRestart || envChanged || entrypointChanged;
+    // Never restart implicitly (issue #449): a restart job is enqueued only
+    // when the caller explicitly asks for one. Saved env/entrypoint changes
+    // are applied by reconfigure-container.js on the next restart.
```
runleveldev, line 647: I know this is doing exactly what the ticket asked for but I don't _fully_ agree with the logic. The *real* problem with the implicit restart is that if, for example, the admin changed the default container vars, that would queue a restart regardless of the change the user made (possibly even if they made no changes). In reality, the implicit restart is good if the user intentionally changes entrypoint/environment because otherwise they may be left to wonder why their changes didn't take effect. Really there's 2 problems here:

1. Invisible (to the user) changes by the admin can make a restart nessecary
2. The user is not properly informed of changes requiring a restart

We can hook into this new solution to fix both of these, but we need one more layer on top. We keep this new "no restarts by default" logic to keep admin changes from bleeding into the container lifecycle behavior BUT if the _user_ makes environment or entrypoint changes (or anything else requiring a restart) then some sort of noticable but unobtrusive UI affordance informs the user that this change requires a restart and the "restart" toggle is switched on. The user can still switch this off manually afterwards but another warning (under the toggle maybe) "this change requires a restart but a restart will not be performed".

runleveldev (changes requested):

cmyers-mieweb:
## Recent changes in commit `e3cb7ad`

### Version is now baked into the client build (review feedback from @runleveldev)
The deployed system has no git checkout, so the runtime git interrogation
(`getVersionInfo`) is gone entirely — `/health` no longer returns `version`
(which also brings it back in line with the OpenAPI schema).

- `package-version` gained a `node` format that emits clean semver
  (`2026.8.1`, prerelease as `-rc1`, snapshots as `+N.ghash` build metadata)
- The packaging build (`make build`, hit by `images/builder` → `make deb`)
  writes that version into the client `package.json` right before
  `vite build`; vite bakes it in as `__APP_VERSION__` — same pattern as
  ozwell-studio
- The footer renders the baked version (linked to GitHub releases) and shows
  "Development build" for dev builds (`0.0.0`)

### Restart toggle follows restart-requiring edits (review feedback from @runleveldev)
Backend stays "no restarts by default" so admin-side changes never bleed into
the container lifecycle, but now:

- Editing env vars or the entrypoint automatically switches the
  "Restart after saving" toggle **on** (compared against the saved container,
  so reverting an edit switches it back off)
- Manually flipping the toggle takes over from the auto behavior
- Toggle off with restart-requiring changes shows an inline warning:
  the change is saved but only applies on the next restart

Also per copilot review: the save toast now uses the backend's `message`, so
update-status wording lives in one place.

runleveldev (approved):

PR #467, opened 2026-09-04
Align git identity with SSSD settings

cmyers-mieweb:
Followup to Issue https://github.com/mieweb/opensource-server/issues/256

Update the container git identity profile to derive LDAP URI, search base, username attribute, and bind credentials from the same SSSD environment used for authentication. Add a workflow smoke test covering bound and anonymous lookups, rootDSE discovery, and no-op cases, and document the first-login git config behavior for LDAP-backed users.


Bobzemob (approved):
LGTM

Issue #471, opened 2026-09-06
Sharing should control SSH access: evaluate owner/collaborators at connect time


horner:
## Problem

The **Sharing** card on a container (owner + collaborators) only controls who sees the container in the manager. It has no effect on SSH.

Today any directory user who matches `SSSD_LDAP_ACCESS_FILTER` (default `(objectClass=*)`) and has an LDAP `sshPublicKey` can SSH into **every** container — shared or not. Sharing and SSH access should be the same thing: not shared ⇒ only the owner can log in; shared ⇒ owner + collaborators.

## Why not fix it in LDAP?

- Every container binds to the same authentik LDAP outpost as `ldap-proxyuser` with identical queries. The directory cannot tell which container is asking, so it cannot answer per-container.
- A per-container `ldap_access_filter` (e.g. `(memberOf=cn=ct-<id>,…)`) still needs a per-container value rendered into `sssd.conf`, which only happens at boot, adds a second source of truth (authentik groups vs the manager DB), and inherits SSSD access-cache latency.

## Approach: evaluate at connect time

sshd already runs a program per login (`AuthorizedKeysCommand`). Use it, plus a PAM `account` hook for the password/MFA path, to ask the manager whether `<user>` is allowed on `<this container>`.

```mermaid
sequenceDiagram
    participant U as ssh client
    participant S as sshd (container)
    participant M as Manager API
    participant L as SSSD / LDAP
    U->>S: connect as alice
    S->>S: AuthorizedKeysCommand ssh-access-authorized-keys alice
    S->>M: GET /api/v1/containers/:id/ssh-access/alice (Bearer container token)
    M-->>S: 204 (owner or collaborator) / 403
    alt allowed
        S->>L: sss_ssh_authorizedkeys alice
        L-->>S: keys
    else denied
        S-->>U: no keys → permission denied
    end
```

- The manager DB (`Container.username` + `ContainerCollaborator`) stays the **single source of truth**.
- Share/unshare takes effect on the **next connection** — no timers, versions, `sshd` reloads.
- Authentication is unchanged: keys still come from the user's LDAP `sshPublicKey`.

### Manager (`create-a-container`)

- Migration: `Containers.sshAccessTokenHash` (argon2, nullable; `null` = not enrolled).
- `middlewares/containerToken.js`: Bearer token → `argon2.verify` against the container's hash.
[...]

PR #472, opened 2026-09-06
feat(ssh): sharing list controls who may SSH into a container

horner:
Part 1 of #471 — the **Sharing** list (owner + collaborators) becomes the SSH allowlist, evaluated by the manager on every connection. Part 2 (enrollment script for existing containers, run per Proxmox node) follows once this is merged.

## How it works

```mermaid
sequenceDiagram
    participant U as ssh client
    participant S as sshd (container)
    participant M as Manager
    participant L as SSSD / LDAP
    U->>S: connect as alice
    S->>S: AuthorizedKeysCommand ssh-access-authorized-keys alice
    S->>M: GET /api/v1/containers/:id/ssh-access/alice (Bearer container token)
    M-->>S: 204 owner/collaborator · 403 otherwise
    alt allowed
        S->>L: sss_ssh_authorizedkeys alice
    else denied
        S-->>U: no keys → permission denied
    end
    S->>M: PAM account: ssh-access-check (same call, covers password/MFA)
```

- Manager DB stays the single source of truth; share/unshare applies on the next connection. No timers, versions, or `sshd` reloads.
- Authentication is unchanged (LDAP `sshPublicKey`); this only adds authorization.
- Manager unreachable ⇒ recent cached allow is honoured (24 h), so the owner keeps access through an outage while strangers stay out.
- Containers without a token (created before this) behave exactly as before and show **SSH not enforced** in the UI.

## Manager (`create-a-container`)

- Migration `Containers.sshAccessTokenHash` (argon2, nullable) + `MANAGER_URL` seeded into default container env vars.
- `Container.sshAllowsUser()`, `rotateSshAccessToken()`, `ensureSshAccessToken(currentLxcEnv)` (reuses the running token on reconfigure), `verifySshAccessToken()`.
- `CONTAINER_ID` / `CONTAINER_SSH_TOKEN` are reserved env keys — stripped from user/admin input, injected by `buildLxcEnvConfig({ sshAccessToken })`.
- `GET /api/v1/containers/:id/ssh-access/:username` — container-token auth; `204` / `403` / `400` bad username / `401` bad token; `Cache-Control: no-store`; denials logged.
- `POST /sites/:siteId/containers/:id/ssh-access/token` — owner/admin mints/rotates (used by the part-2 enrollment script).
- Serializer: `sshAccessEnforced`.

## Base image (`images/base`)

- `ssh-access-check` — one script, two callers: `AuthorizedKeysCommand` wrapper (`AuthorizedKeysCommandUser sshaccess`) and `pam_exec` in the account phase. Token is passed to curl via stdin config, never argv.
- `ssh-access-setup.service` — copies the three env vars from PID 1 into `/etc/ssh-access/` (0640 root:sshaccess); `environment.sh` filters `CONTAINER_SSH_TOKEN` out of `/etc/environment`.

## Client

Sharing copy now says what it does; `SshAccessBadge` shows **SSH enforced** / **SSH not enforced**.

## Tests

- `routers/api/v1/__tests__/ssh-access.api.test.js` — owner/collaborator 204, stranger 403, unshare → 403, invalid usernames 400, missing/wrong/rotated/user-API-key tokens 401, unenrolled 401, mint permissions (collaborator 403, stranger 404), reserved keys not overridable, token reuse/rotation.
- `images/base/test-ssh-access-check.sh` — 19 checks with a stubbed `curl` (allow/deny/cache/outage/401/invalid user/PAM_USER).
- `npx jest`: 118 pass; the one failure (`mcp-proxy › not mounted when no MCP server is configured`) is pre-existing on `main`.

## Deploy notes

1. Run migrations; set **Settings → Default Container Environment Variables → `MANAGER_URL`** to the manager URL reachable from containers.
2. Rebuild/publish the base image.
3. New containers are enforced automatically. Existing containers: part 2.

## Out of scope

Killing live sessions on unshare; `ldapusers` passwordless sudo (sharing = admin-level trust); Proxmox ACLs for collaborators.


images/base/Dockerfile
```diff
@@ -32,6 +32,18 @@ COPY --chmod=0440 ldapusers /etc/sudoers.d/ldapusers
...
+# The sshaccess user exists so the token file need not be readable by nobody.
+COPY --chmod=0755 ssh-access-check.sh /usr/local/bin/ssh-access-check
+COPY --chmod=0755 ssh-access-authorized-keys.sh /usr/local/bin/ssh-access-authorized-keys
+COPY --chmod=0755 ssh-access-setup.sh /usr/local/bin/ssh-access-setup.sh
+COPY ssh-access-setup.service /etc/systemd/system/ssh-access-setup.service
+RUN useradd --system --no-create-home --shell /usr/sbin/nologin sshaccess && \
+    sed -i '/^@include common-account/i account required pam_exec.so quiet /usr/local/bin/ssh-access-check' /etc/pam.d/sshd && \
+    systemctl enable ssh-access-setup.service
+
 # The following service ensures environment variables set in the OCI metadata
```
runleveldev: **Consolidate to a single PAM `account` hook; drop the AuthorizedKeysCommand wrapper.**

The design currently enforces authorization at *two* points for the same decision:

1. `AuthorizedKeysCommand` wrapper (`ssh-access-authorized-keys.sh`) — gates the key path
2. this `pam_exec` `account` hook — gates the password/MFA path

But the base image doesn't override `UsePAM`, so Debian's default `UsePAM yes` is in effect. That means the PAM `account` phase runs on **every** successful authentication, including publickey — not just password/MFA. So the `account` hook alone already covers both paths. The PR's own `test-ssh-access-check.sh` demonstrates this (the `PAM_USER honoured` case is the key path being gated by the same script).

Concretely, I'd suggest:

- **Keep** this `account required pam_exec.so … ssh-access-check` line — it's the single, idiomatic choke point. `account` is exactly where "may this authenticated user log into this host" belongs, and it's method-agnostic.
- **Revert** `50-sss-ssh-authorizedkeys.conf` back to stock `AuthorizedKeysCommand /usr/bin/sss_ssh_authorizedkeys %u` with `AuthorizedKeysCommandUser nobody`. Then `ssh-access-authorized-keys.sh` can be deleted entirely.

Benefits:
- One enforcement path instead of two → half the surface to reason about, test, and keep in sync.
- No `curl` on every key negotiation. Today a denied user still triggers a manager round-trip during `AuthorizedKeysCommand` *and* the LDAP key fetch; with account-only, the key exchange stays fully native (`sss_ssh_authorizedkeys` unchanged) and the single callback happens at `account`.
- Restores the stock, well-understood `AuthorizedKeysCommand` rather than a shell wrapper that shells out to `curl` mid-handshake.

One thing to confirm if you take this: a denied user will now fail at `account` *after* key/password auth succeeds, so they'll see `Permission denied` at the account stage rather than "no keys offered." That's standard PAM behaviour and arguably clearer, but worth a note in the docs since the current copy describes "no keys → permission denied."

If there's a reason the key path must be denied *before* `account` (e.g. to avoid emitting any auth-success signal), that's the tradeoff to call out explicitly — but for owner/collaborator gating I don't see one.

images/base/50-sss-ssh-authorizedkeys.conf
```diff
@@ -1,4 +1,7 @@
 # Fetch users' authorized SSH public keys from SSSD (which serves the LDAP
-# `sshPublicKey` attribute). %u is replaced by sshd with the login username.
-AuthorizedKeysCommand /usr/bin/sss_ssh_authorizedkeys %u
-AuthorizedKeysCommandUser nobody
+# `sshPublicKey` attribute) — but only after the manager confirms the user is
+# this container's owner or a collaborator (see ssh-access-check.sh). %u is
+# replaced by sshd with the login username. Runs as the sshaccess user so it
```
runleveldev: If you adopt the single-PAM-hook approach (see my comment on the Dockerfile), this file should revert to stock:

```
AuthorizedKeysCommand /usr/bin/sss_ssh_authorizedkeys %u
AuthorizedKeysCommandUser nobody
```

and `ssh-access-authorized-keys.sh` can be removed. The `account` PAM hook already gates the key path because `UsePAM yes` (Debian default, not overridden here) runs the `account` phase for publickey logins too. Keeping the wrapper here means a `curl` round-trip on every key negotiation and a second copy of the enforcement logic to maintain.

A side benefit: reverting this also removes the need for the `sshaccess` user to be the `AuthorizedKeysCommandUser`. The only remaining reason for the `sshaccess` user is readable ownership of `/etc/ssh-access/token` for the `pam_exec` script, which runs as root anyway — so you may be able to drop the dedicated user entirely and keep the token `0600 root:root`. Worth checking whether `pam_exec` (root) is the sole reader once the wrapper is gone.

runleveldev (commented):
## Suggestion: single native PAM `account` hook instead of two enforcement points

I looked at whether there's a more Linux/LDAP-native way to do this before reviewing correctness. Two findings:

**On the LDAP-group route** (per-container authentik group + `ldap_access_filter (memberOf=…)`): I think your decision to *not* go this way is justified. The authentik LDAP outpost is read-only (memberships would have to be written via the authentik REST API, which the manager has no client for today), it adds a second source of truth to keep in sync, and it inherits SSSD access-cache latency. Keeping the manager DB authoritative and evaluating at connect time is reasonable. (Note: the outpost *does* expose `memberOf`, so the filter would technically work — the objection is complexity/consistency, not feasibility.)

**On the enforcement mechanism** — this is where I'd push for a change. The PR gates the same decision in two places (AuthorizedKeysCommand wrapper for the key path, `pam_exec account` for the password path). Because `UsePAM yes` is the default and isn't overridden, the PAM `account` phase already runs for publickey logins too, so a **single `account` hook covers both paths**. Reverting the AuthorizedKeysCommand to stock `sss_ssh_authorizedkeys` and keeping only the `account` hook is more native, removes a `curl` from every key negotiation, and halves the logic to maintain. Details in the two line comments.

This is a design suggestion, not a blocker — happy to discuss the one tradeoff I noted (denial surfaces at `account` after auth succeeds rather than as "no keys offered").

runleveldev:
Applied the single-PAM-`account`-hook simplification in 25ce0417.

- `AuthorizedKeysCommand` reverted to stock `sss_ssh_authorizedkeys` (as `nobody`); `ssh-access-authorized-keys.sh` removed.
- Enforcement is now the one `pam_exec` `account` hook, which covers publickey and password/MFA alike since `UsePAM yes` runs the account phase on every successful auth.
- With the wrapper gone, `ssh-access-check` runs only as root via `pam_exec`, so the dedicated `sshaccess` user is dropped and `/etc/ssh-access/` + the token are root-only (`0700` dir, `0600` token).

Removes a manager round-trip on every key negotiation and a second copy of the enforcement logic. `test-ssh-access-check.sh` still passes 19/19 (incl. the `PAM_USER` path); PR description/diagram updated to match.

